PR #6, opened 2019-05-19
Connecting Frontend to Backend API

davish:
This PR updates frontend API endpoints and expected JSON schema so that the new, frontend can query from and display information from the Django REST framework backend. This touches a lot of files, on both frontend and backend.

The rationale behind changing the schema was that generally, the old schema was pretty esoteric. As an example, the ending of a meeting block could be found with `startHour + hourLength`, which doesn't make too much semantic sense.

Still to-do after this PR:
- Display crosslistings
- Add requirements to API response
- Add PCR ratings to courses
- Generally add more info on courses to API

### DIFF
--- a/frontend/plan/src/components/schedule/Schedule.js
+++ b/frontend/plan/src/components/schedule/Schedule.js
@@ -3,7 +3,7 @@ import PropTypes from "prop-types";
 
 import connect from "react-redux/es/connect/connect";
 
-import { removeSchedItem } from "../../actions";
+import { removeSchedItem, fetchSectionInfo } from "../../actions";
 
 import "./schedule.css";
 import Days from "./Days";
@@ -80,7 +80,7 @@ export const getConflictGroups = (meetings) => {
 
 class Schedule extends Component {
     render() {
-        const { schedData, removeSection } = this.props;
+        const { schedData, removeSection, focusSection } = this.props;
         const sections = schedData.meetings || [];
 
         if (sections.length < 1) {
@@ -115,7 +115,7 @@ class Schedule extends Component {
                     // if we've used all the colors, it's acceptable to start reusing colors.
                     used = [];
                 }
-                let i = hashString(c);
+                let i = Math.abs(hashString(c));
                 while (used.indexOf(colors[i % colors.length]) !== -1) {
                     i += 1;
                 }
@@ -139,9 +139,10 @@ class Schedule extends Component {
                     course: {
                         color,
                         id: s.id,
-                        coreqFulfilled: s.associated_sections.filter(
-                            coreq => sectionIds.indexOf(coreq.id) !== -1
-                        ).length > 0,
+                        coreqFulfilled: s.associated_sections.length === 0 ||
+                            s.associated_sections.filter(
+                                coreq => sectionIds.indexOf(coreq.id) !== -1
+                            ).length > 0,
                     },
                     style: {
                         width: "100%",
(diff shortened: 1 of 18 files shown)


frontend/plan/src/components/selector/SectionDisplay.js
```diff
@@ -1,23 +1,63 @@
...
 
+import Badge from "../Badge";
+
 export default class SectionDisplay extends Component {
     stripTime = (s) => {
         let newS = s.replace(" to ", "-");
         newS = newS.replace("on", "");
         return newS;
-    }
+    };
```
eyingxuan: I understand that this is how we get the data from the Penn Registrar API, but I was wondering if it would be better if this was handled in the backend instead.

frontend/plan/src/components/selector/SectionDisplay.js
```diff
@@ -1,23 +1,63 @@
...
+
+    getTimeString = (meetings) => {
```
eyingxuan: I understand that this is how we get the data from the Penn Registrar API, but I was wondering if it would be better if this was handled in the backend instead.

eyingxuan:
With the introduction of `fetch()` directly in `fetchCourseSearch()` and `fetchSectionInfo()` in `actions/index.js` I think we can remove the `requestSearch` and `requestSectionInfo` functions from our actions as I don't see them anywhere in our reducers. I don't think they have to be called either in line 189 and line 202 in `actions/index.js`

frontend/plan/src/components/schedule/Schedule.js
```diff
@@ -228,6 +235,7 @@ const mapStateToProps = state => (
 const mapDispatchToProps = dispatch => (
     {
         removeSection: idDashed => dispatch(removeSchedItem(idDashed)),
+        focusSection: id => dispatch(fetchSectionInfo({ param: id })),
```
eyingxuan: I understand that requirements and associated section have yet to be implemented in the backend, but just wanted to point out that we might want to think about how to separate the information that users get from clicking on a course versus the information that users get from clicking on a section. Right now the same information is displayed when clicking on a course from `search_results.js` and any one of the blocks, while nothing is shown when clicking on a section in the `SectionList` component.

eyingxuan:
We might also want to consider either removing course ID hyphenation directly in the backend, or auto-adding hyphenation for users who search for "CIS 120" instead of "CIS-120" in `fetchCourseSearch`

davish:
> We might also want to consider either removing course ID hyphenation directly in the backend, or auto-adding hyphenation for users who search for "CIS 120" instead of "CIS-120" in `fetchCourseSearch()`

I think the right way to do this would be to remove the dash in the backend. In reality, any of these should produce the same result: `cis120`, `Cis120`, `cis 120`, `CIS-120`.

frontend/plan/src/components/selector/SectionDisplay.js
davish: Since the API will eventually be used for more than just PCP, I feel like we should try to keep most of the product-specific computation on the frontend. I do agree that the `SectionDisplay` component is getting pretty crowded, and there's probably an opportunity to refactor all of these `meeting` time utility functions, like getTimeString and finding conflicts, into a separate file.

frontend/plan/src/components/schedule/Schedule.js
davish: That's a really good point!

frontend/plan/src/components/selector/SectionDisplay.js
eyingxuan: On second thought, that's actually the better move you're right. Especially since we do use the meeting times data in its current form from the API to generate the blocks in the schedule.

Issue #8, opened 2019-05-20
"Search as you type" feature


eyingxuan:
Users have to press enter in order to search for a course. This can be changed by adding a state in the SearchBar component storing the current course ID typed and implementing a function to pass as an onChange prop to the input field.

eyingxuan:
Implemented at #10

PR #10, opened 2019-05-21
Added search as you type functionality

eyingxuan:
This PR implements #8. Users no longer need to press enter to search after typing a course ID.


eyingxuan:
That's a good point. I can make it so that we only query when the user stops typing.

eyingxuan:
Yeap! LGTM

PR #13, opened 2019-05-25
Search Bar and Filters UI

eyingxuan:
This PR implements the filter functionality as well as the entire search bar according to the design mockups. React Hooks have been introduced in this PR. 

- [x] Refactored OutClickable 
- [x] Implement search bar based on design mockups


eyingxuan:
I think we should consider merging this PR first before I work on the logic/connecting to the backend in another PR. I would like to introduce redux-saga in order to handle fetch latencies (people switching around filters too quickly, slow internet speeds) and redux-saga allows for cancelling past actions, which will be very useful.

davish:
I think splitting up the PR's is a good idea. Just wondering, though, what's the difference between `redux-saga` and `redux-thunk` (which is already installed/used for searches)? I was having trouble finding stuff in their documentation.

eyingxuan:
`redux-saga` has more features, and the one I'm looking for particularly that we can't achieve with `redux-thunk` is to cancel an action. For instance, if the user selects a filter and before the results and returned by the server, they trigger another request by typing in the search bar or apply another filter, we can cancel the previous fetch request and take only the newest one.

davish:
Ah, got it. Would it be worth replacing thunk entirely?

eyingxuan:
Yea, I actually think thunk and saga don't work well together. Replacement shouldn't take too long though, AFAIK only two actions uses thunk right now so rewriting them should be fairly quick.

PR #23, opened 2019-09-21
Lint

ArmaanT:
This PR sets up linting and code coverage as well as fixing existing code.


alert/views.py
```diff
@@ -1,15 +1,16 @@
...
+from django.http import Http404, HttpResponse, HttpResponseRedirect, JsonResponse
 from django.shortcuts import render
-from django.http import HttpResponseRedirect, JsonResponse, Http404, HttpResponse
 from django.urls import reverse
 from django.views.decorators.csrf import csrf_exempt
 
-from .models import *
-from .tasks import generate_course_json, send_course_alerts
-from options.models import get_bool
+from alert.models import Registration, RegStatus, record_update, register_for_course, update_course_from_record
```
davish: I'll be sad that lint doesn't like relative imports or `*` imports but it's probably for the better.

alert/views.py
```diff
@@ -129,9 +130,9 @@ def accept_webhook(request):
 
     if username != settings.WEBHOOK_USERNAME or \
             password != settings.WEBHOOK_PASSWORD:
-        return HttpResponse('''Your credentials cannot be verified. 
-        They should be placed in the header as &quot;Authorization-Bearer&quot;,  
-        YOUR_APP_ID and &quot;Authorization-Token&quot; , YOUR_TOKEN"''', status=401)
+        return HttpResponse("""Your credentials cannot be verified.
```
davish: Should be a separate PR but need a `TODO` to change this error message, since we're actually using basic auth for the webhook

PR #25, opened 2019-09-24
Third Party Integration

davish:
This PR adds models for API keys and a route to register users for alerts with a JSON api.


courses/models.py
```diff
@@ -287,3 +288,30 @@ def satisfying_courses(self):
...
+
+
+"""
+3rd-Party API
+"""
+
+PCA_REGISTRATION = 'PCA_REGISTRATION'
+
+
+class APIPrivelege(models.Model):
```
ArmaanT: Typo here. APIPrivelege -> APIPrivilege. This would need to be changed everywhere.

courses/models.py
```diff
@@ -287,3 +288,30 @@ def satisfying_courses(self):
...
+    """
+    Describes a type of access privelege that an API key can have.
+    """
+    code = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True)
+
+
+class APIKey(models.Model):
```
ArmaanT: Does it make more sense to build our own API key, or just use an existing package like [djangorestframework-api-key](https://github.com/florimondmanca/djangorestframework-api-key)?

alert/tests.py
```diff
@@ -399,6 +402,95 @@ def test_wrong_user(self, mock_alert):
         self.assertEqual(0, CourseUpdate.objects.count())
 
 
+@override_settings(SWITCHBOARD_TEST_APP='pca')
+class APIRegisterTestCase(TestCase):
+    def setUp(self):
+        set_semester()
+        self.client = Client()
+        self.course, self.section = get_course_and_section('CIS-120-001', TEST_SEMESTER)
+        self.permission = APIPrivelege.objects.get_or_create(code=PCA_REGISTRATION)[0]
```
ArmaanT: I know this is a test case and it doesn't really matter, but I think you should just use `create` not `get_or_create` because the privilege can't exist unless you define it in the test case.

alert/models.py
```diff
@@ -112,11 +124,11 @@ def resubscribe(self):
         return new_registration
 
 
-def register_for_course(course_code, email_address, phone):
+def register_for_course(course_code, email_address, phone, source='PCA', api_key=None):
```
ArmaanT: I think it's generally a best practice to define `"PCA"` as a constant and refer to that instead of using the string itself.

alert/tests.py
```diff
@@ -76,6 +77,8 @@ def test_successful_registration(self):
         self.assertEqual('e@example.com', r.email)
         self.assertEqual('+15555555555', r.phone)
         self.assertFalse(r.notification_sent)
+        self.assertEqual('PCA', r.source)
```
ArmaanT: Same potential issue here with the constant.

courses/models.py
davish: This looks like it makes sense. I'll see if I have time to plug this in tonight.

davish:
[3rd-party integrations](https://trello.com/c/ZeAjpy0M/59-3rd-party-integrations)

courses/models.py
davish: After talking in person, Armaan and I came to the agreement that the linked library 1) doesn't implement a lot of functionality that we couldn't easily implement ourselves and 2) doesn't seem super popular and might be at risk of abandonment. Reason 2 tends to outweigh the concerns from reason 1 of reinventing the wheel in this instance.

PR #28, opened 2019-10-01
Add under construction site for PCP/PCS

davish:
This PR is step 1 in shutting down the current version of PCS. We'll redirect to an "under construction" page similar to Penn Clubs:

Screen Shot 2019-10-01 at 9 36 54 AM


eyingxuan:
Frontend lgtm! Not too sure how Django works, but should we comment out the other paths after the root url match in urls.py?

davish:
> Frontend lgtm! Not too sure how Django works, but should we comment out the other paths after the root url match in urls.py?

I think it should be fine. Only thing is that means that people snooping around will be able to see our representation of course data, which they could access anyway through `api.pennlabs.org`.

PR #30, opened 2019-10-06
Fix meeting duplicates

davish:
Change how `set_meetings` works so there aren't duplicates when you load in courses with rooms.


davish:
[Fix duplicate courses from adding new data](https://trello.com/c/3RwycrjU/52-fix-duplicate-courses-from-adding-new-data)

courses/util.py
```diff
@@ -65,17 +65,19 @@ def get_room(building_code, room_number):
     return room
 
 
-def add_meetings(section, meetings):
+def set_meetings(section, meetings):
+    for meeting in section.meetings.all():
+        meeting.delete()
```
ArmaanT: You should be able to replace this with something like `section.meetings.all().delete()`

PR #32, opened 2019-10-08
Command-Line Scripts for loading in PCR Data

davish:
This PR adds an integration with the PCR API that allows Penn Courses to sync PCR data. Because the data doesn't change very often (once per semester) and it takes a really long time to get from the API, we shouldn't be hitting the PCR API very often. The solution to this is that there's one command `pcrbackup` which hits the PCR API and outputs data to a directory of JSON files which serve as a backup of all of PCR data. The command `loadreviews` then loads in all the reviews from the JSON files in the specified.

In practice, developers will have access to the directory generated by `pcrbackup` and won't have to run it themselves.

An improvement might be to take in a `.tar.gz` file instead of a directory.


davish:
[Load in PCR data](https://trello.com/c/SSW9NuZx/37-load-in-pcr-data)

PR #34, opened 2019-10-11
Course Filters

davish:
This PR adds in more filters for the filter bar for PCP. Documentation can be found on Notion here: https://www.notion.so/pennlabs/API-Specification-2ae3ab398eb54475a1efe6e530dee17d#bdce343ce6d24a319a7b1abc01c249fc

### DIFF
--- a/plan/views.py
+++ b/plan/views.py
@@ -5,10 +5,11 @@ from courses.views import CourseDetail, CourseList
 
 from .search import TypedSearchBackend
 from .serializers import CourseDetailWithReviewSerializer, CourseListWithReviewSerializer
+from .filters import requirement_filter
 
 
 class CourseListSearch(CourseList):
-    filter_backends = (TypedSearchBackend, )
+    filter_backends = [TypedSearchBackend]
     search_fields = ('full_code', 'title', 'sections__instructors__name')
     serializer_class = CourseListWithReviewSerializer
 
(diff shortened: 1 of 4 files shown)


plan/views.py
```diff
@@ -1,33 +1,32 @@
-from django.db.models import Q
-
-from courses.models import Requirement
 from courses.views import CourseDetail, CourseList
 
+from .filters import bound_filter, requirement_filter
 from .search import TypedSearchBackend
 from .serializers import CourseDetailWithReviewSerializer, CourseListWithReviewSerializer
```
ArmaanT: No relative imports please

plan/views.py
```diff
@@ -1,33 +1,32 @@
...
+            'cu': bound_filter('sections__credits'),
+            'course_quality': bound_filter('course_quality'),
+            'instructor_quality': bound_filter('instructor_quality'),
+            'difficulty': bound_filter('difficulty')
+        }
+
+        for field, filter_func in filters.items():
+            param = self.request.query_params.get(field)
+            if param is not None:
+                queryset = filter_func(queryset, param, self.get_semester())
```
ArmaanT: Does it make more sense to move this logic to the filter method itself? I haven't really looked into filtering all that much, so it's possible I'm completely wrong. [Docs](https://www.django-rest-framework.org/api-guide/filtering) on DRF filtering.

davish: From the docs it seemed like the right place to do non-generic filtering is in the `get_queryset()` method on the view, since that's where you can get access to query parameters and stuff.

plan/views.py
davish: is that a rule we could add to flake8?

ArmaanT: There doesn't seem to be an easy way to do that in either `flake8` or `isort`, but I'll keep looking

plan/views.py
ArmaanT: Makes sense to me. Resolved!

PR #46, opened 2019-10-18
Implement Accounts

mureytasroc:
I just installed django-labs-accounts and tested out the login sequence (it works).  So I think we just have to add the appropriate links on the frontend and utilize the sessionid cookie that is created as a result of the login sequence.


PennCourses/settings/base.py
```diff
@@ -56,6 +57,18 @@
...
 
+#From labs-accounts
+AUTHENTICATION_BACKENDS = [
+    'accounts.backends.LabsUserBackend',
+    'django.contrib.auth.backends.ModelBackend',
+]
+
+#From labs-accounts
+PLATFORM_ACCOUNTS = {
+    'REDIRECT_URI': os.environ.get('LABS_REDIRECT_URI', 'https://penncourseplan.com/accounts/callback/'),
```
ArmaanT: Hmm, we have a weird issue where this one instance of Django is serving PCA and PCP. I'm not a huge fan of just hardcoding the URI to be for PCP because then PCA logins will route though PCP, but I'm also not sure of a good alternative. @davish thoughts?

PennCourses/settings/base.py
```diff
@@ -56,6 +57,18 @@
     'review',
 ]
 
+#From labs-accounts
+AUTHENTICATION_BACKENDS = [
+    'accounts.backends.LabsUserBackend',
+    'django.contrib.auth.backends.ModelBackend',
```
davish: I think eventually we'll want to remove the ModelBackend, but it's a good idea rn to have a back-up if PennKeys fail for some reason

PennCourses/settings/base.py
davish: We haven't set up `api.penncourses.org` yet, but this might be a good way to do it. Although I think for the PCP launch it's fine to send it thru PCP, since PCA isn't using accounts yet.

davish: Also though, we would also want this to be different in dev and prod, right? redirect to `localhost/accounts/callback/` in dev? or does it not work like that

mureytasroc: You can change the env variable to set it to localhost in dev

mureytasroc: So eventually we should change that uri to api.penncourses.org?

PennCourses/settings/base.py
mureytasroc: Yeah I just added it because it was in the django-labs-accounts installation instructions but that makes sense

PennCourses/settings/base.py
mureytasroc: Ok I squashed the commits.

ArmaanT: @davish I'll finalize some of the details of using accounts in development and add the configuration to the django template and here.

davish:
[Add Accounts](https://trello.com/c/vkZ6tHYV/51-add-accounts)

PR #54, opened 2019-10-19
Schedules backend

mureytasroc:
This will probably need a lot of changes....


plan/models.py
```diff
@@ -1 +1,34 @@
...
+
+from courses.models import Section
+
+
+class Schedule(models.Model):
+    """
+    Used to save schedules created by users on PCP
+    """
+    person = models.ForeignKey(get_user_model(), on_delete=models.CASCADE)
+
```
davish: Can we take out these blank lines between fields?

plan/serializers.py
```diff
@@ -88,28 +89,26 @@ class Meta:
...
 
     @staticmethod
     def setup_eager_loading(queryset):
         queryset = annotations.sections_with_reviews()
-        queryset = queryset.prefetch_related(
-            'course__department',
-            'meetings__room__building',
-            'associated_sections__course__department',
-        )
+        # queryset = queryset.prefetch_related(
```
davish: Should this be commented out?

plan/tests.py
```diff
@@ -193,3 +197,123 @@ def test_filter_courses_by_review_excluded(self):
...
+                                    content_type='application/json')
+        self.assertEqual(response.status_code, 201)
+        response = self.client.get('/schedules/')
+        self.assertEqual(200, response.status_code)
+        self.assertEqual(2, len(response.data))
+        self.assertEqual(response.data[1]['name'], 'New Test Schedule')
+        self.assertEqual(len(response.data[1]['sections']), 2)
+        self.assertEqual(response.data[1]['sections'][0]['id'], 'CIS-121-001')
+
+    # vvv implement below tests
```
davish: Can remove this comment

plan/views.py
```diff
@@ -32,3 +38,67 @@ def get_queryset(self):
...
+    return sections
+
+
+class ScheduleViewSet(viewsets.ModelViewSet):
+    serializer_class = ScheduleSerializer
+    http_method_names = ['get', 'post', 'delete', 'put']
+    permission_classes = [IsAuthenticated]
+
+    def update(self, request, pk=None):
+        existing_obs = Schedule.objects.filter(id=pk)
```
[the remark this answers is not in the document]
davish: You can then catch the Schedule.DoesNotExist exception, which is thrown if you call get and an object doesn't exist.

plan/views.py
```diff
@@ -32,3 +39,83 @@ def get_queryset(self):
...
+        if 'semester' not in request.data:
+            request.data['semester'] = get_value('SEMESTER', None)
+
+        for s in request.data['sections']:
+            if s.get('semester') != request.data.get('semester'):
+                return Response({'detail': 'Semester uniformity invariant violated.'},
+                                status=status.HTTP_400_BAD_REQUEST)
+
+        try:
+            existing_obs.update(person=request.user,
```
davish: `existing_obs` and `ob` are pretty undescriptive variable names, especially because they refer to almost the same thing?

plan/views.py
```diff
@@ -32,3 +39,83 @@ def get_queryset(self):
...
+                                semester=request.data.get('semester'),
+                                name=request.data.get('name'),
+                                )
+            ob = existing_obs[0]
+            ob.sections.set(get_sections(request.data))
+            s = ScheduleSerializer(ob)
```
davish: again, `s` is a kind of confusing variable name. maybe `serialized_schedule`?

plan/views.py
```diff
@@ -32,3 +39,83 @@ def get_queryset(self):
...
+            return Response(s.data, status=status.HTTP_202_ACCEPTED)
+        except IntegrityError:
+            return Response({'detail': 'Unique constraint violated'}, status=status.HTTP_400_BAD_REQUEST)
+
+    def create(self, request, *args, **kwargs):
+        existing_obs = Schedule.objects.filter(id=request.data.get('id'))
```
davish: Lots of the comments are going to be similar from above.

plan/views.py
```diff
@@ -32,3 +39,83 @@ def get_queryset(self):
...
+        if 'semester' not in request.data:
+            request.data['semester'] = get_value('SEMESTER', None)
+
+        for s in request.data['sections']:
+            if s.get('semester') != request.data.get('semester'):
+                return Response({'detail': 'Semester uniformity invariant violated.'},
+                                status=status.HTTP_400_BAD_REQUEST)
+
+        try:
+            if 'id' in request.data:
```
davish: The correct behavior is probably to ignore the given id when creating a schedule, since that might cause a collision with an existing schedule that we wouldn't want to happen.

PR #96, opened 2019-10-20
fixed time stats

szarkesh:


frontend/plan/src/components/schedule/Stats.js
```diff
@@ -57,22 +51,22 @@ class Stats extends Component {
             if (str) {
                 const course = str.substring(0, str.indexOf("-", str.indexOf("-") + 1)); // finds course (irrespective of section)
                 if (course in courseDifficulties) {
-                    courseDifficulties[course] += (section.difficulty ? section.difficulty : 2.0);
+                    courseDifficulties[course] += (section.difficulty ? section.difficulty : 2.5);
                     courseWorkloads[course] += (section.work_required
-                        ? section.work_required : 2.0);
```
davish: I think the better way to do this would be to not add to the sum when the field isn't there? From some back of the napkin math, doesn't seem like either 2.0 or 2.5 does what we want it to do (keep the avg the same).

frontend/plan/src/components/schedule/Stats.js
```diff
@@ -57,22 +51,22 @@ class Stats extends Component {
...
+                        ? section.instructor_quality : 2.5);
                     courseQualities[course] += (section.course_quality
-                        ? section.course_quality : 2.0);
+                        ? section.course_quality : 2.5);
                     courseCUs[course] += (section.credits ? section.credits : 1);
                     courseRepeats[course] += 1;
                 } else {
-                    courseDifficulties[course] = (section.difficulty ? section.difficulty : 2.0);
-                    courseWorkloads[course] = (section.work_required ? section.work_required : 2.0);
+                    courseDifficulties[course] = (section.difficulty ? section.difficulty : 2.5);
```
davish: could we generalize a lot of this code? it seems like a lot is copy/paste and for maintainability I think it would be good to abstract these if statements and other logic out into functions, and iterate through with for loops.

frontend/plan/src/components/schedule/Stats.js
```diff
@@ -143,49 +137,45 @@ class Stats extends Component {
                         {" "}
                     </div>
                 </div>
-                { false
```
davish: Could we generalize these into two components for the gauge and the time stats? Seems like a lot of copy/paste here too.

PR #97, opened 2019-10-22
repaced min hours / day with credits

szarkesh:


frontend/plan/src/components/schedule/Stats.js
```diff
@@ -32,15 +32,15 @@ class Stats extends Component {
...
+        let totalCUs;
         let maxHoursADay;
         let totalHours;
         let averageHours;
         let avgs = {};
         if (meetings.length === 0) {
             earliestStart = "—";
             latestEnd = "—";
-            minHoursADay = "—";
+            totalCUs = "–";
```
davish: I think this should be an em dash (option-dash) instead of a regular dash:
Screen Shot 2019-10-22 at 1 40 28 PM

PR #100, opened 2019-10-24
Collapsible requirements

dlike230:
Instead of having a popover for requirements, requirements can now be expanded and collapsed. The design needs to be improved before this is merged.

Collapsed (default state):
image

Expanded:
image


dlike230:
@davish ready to merge

PR #101, opened 2019-10-27
Improved Modal UX

dlike230:
Changes:
- Default value for rename schedule is the old schedule name
- Default value for create schedule is "schedule name"
- Clicking out of an empty input returns it to its default value

### DIFF
--- a/frontend/plan/src/components/modals/NameScheduleModalInterior.js
+++ b/frontend/plan/src/components/modals/NameScheduleModalInterior.js
@@ -32,7 +32,7 @@ const NameScheduleModalInterior = ({
                 value={userInput}
                 type="text"
                 ref={ref => setInputRef(ref)}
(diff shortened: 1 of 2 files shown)


frontend/plan/src/components/modals/NameScheduleModalInterior.js
```diff
@@ -1,13 +1,24 @@
...
+        const listener = (event) => {
+            if (!userInput && inputRef && !inputRef.contains(event.target)) {
+                setUserInput(defaultValue);
+            }
+        };
+        document.addEventListener("click", listener);
+        return () => {
+            document.removeEventListener("click", listener);
+        };
+    });
```
davish: refactor to use Ying's OutClickable hook

dlike230: Ying's hook doesn't work here because I have to use the ref for something else too.

PR #102, opened 2019-10-27
Introduced Checkbox Filters (Type, CU Filters)

eyingxuan:
~~- [ ] Make sure redux logic handles change in CU data type before deploying to master!~~


frontend/plan/src/actions/index.js
```diff
@@ -219,7 +220,7 @@ function buildCourseSearchUrl(filterData) {
     }
 
     // Range filters
-    const filterFields = ["difficulty", "course_quality", "instructor_quality", "cu"];
+    const filterFields = ["difficulty", "course_quality", "instructor_quality"];
     const defaultFilters = [[0, 4], [0, 4], [0, 4], [0.5, 2]];
```
eyingxuan: Should remove the 4th element

PR #103, opened 2019-10-27
Touchups shaya

szarkesh:


dlike230:
@szarkesh It looks like all of these changes are changes I've seen in your mobile pull request.  Can you confirm if this is the case so I can close/review this PR accordingly?

PR #107, opened 2019-10-30
Mobile version

szarkesh:
Added fully-functional mobile version with tabs at the top

image

image


frontend/plan/src/components/schedule/MyCircularProgressBar.js
```diff
@@ -7,7 +7,7 @@ export default function MyCircularProgressBar(props) {
     let color;
     const { value } = props;
     switch (true) {
-        case value === 0:
+        case value === 0 || isNaN(value):
```
dlike230: I think we should try to avoid dealing with NaNs. Maybe make the value -1 if it's invalid?

szarkesh: done

frontend/plan/src/components/search/SearchBar.js
```diff
@@ -71,6 +75,84 @@ function SearchBar({
...
+                                rangeProperty="instructor_quality"
+                                isDisabled={isLoading}
+                            />
+                        </DropdownButton>
+                    </div>
+                )}
+            </div>
+        );
+    }
+
```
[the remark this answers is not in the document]
szarkesh: there's some functionality difference as well, since the filters only show up on mobile on button press -- also, the button for filtering is different, and the sizing and spacing of the elements is so different that I feel like it would be more confusing using CSS.

but I did abstract the repeated code (dropdowns) so that it's not written twice

frontend/plan/src/styles/App.css
```diff
@@ -491,7 +491,7 @@ ul {
 
 @media all and (max-width: 480px){
   .box{
-    min-height: 500px;
+    min-height: calc(100vh - 8em);
```
dlike230: Try to avoid using calc -- it's buggy and not well supported. For this pull request though we can keep it in since the min height requirement isn't very important.

PR #108, opened 2019-11-01
Accounts frontend

dlike230:


frontend/plan/src/actions/index.js
```diff
@@ -191,7 +195,7 @@ export const loadRequirements = () => (
                         console.log(error);
                     }),
                 (error) => {
-                // eslint-disable-next-line no-console
+                    // eslint-disable-next-line no-console
                     console.log(error);
```
davish: I know this isn't related to this PR, but we should probably remove this log statement, right?

frontend/plan/src/reducers/schedule.js
```diff
@@ -89,6 +89,57 @@ const nextAvailable = (scheduleName, used) => {
...
+        case UPDATE_SCHEDULES:
+            // eslint-disable-next-line
+            const { schedulesFromBackend } = action;
+            // eslint-disable-next-line
+            const newScheduleObject = { ...state.schedules };
+            // eslint-disable-next-line
+            const newCart = [...state.cartSections];
+            if (schedulesFromBackend) {
+                schedulesFromBackend.forEach(({ id: scheduleId, title, sections, semester }) => {
+                    if (title === "cart") {
```
davish: Might want some documentation here about how the cart is stored as a schedule on the backend.

frontend/plan/src/reducers/schedule.js
```diff
@@ -116,7 +228,15 @@ export const schedule = (state = initialState, action) => {
                 schedules: {
                     ...state.schedules,
                     [nextAvailable(action.scheduleName, state.schedules)]:
-                        state.schedules[action.scheduleName],
+                        {
+                            ...withoutId(state.schedules[action.scheduleName]),
+                            pushedToBackend: false,
+                            updated_at: Date.now(),
+                            isNew: {
```
davish: Why is `isNew` an object here instead of a boolean?

frontend/plan/src/syncutils.js
```diff
@@ -0,0 +1,101 @@
...
+                    let schedName = Object.keys(scheduleStateInit.schedules)
+                        .reduce((acc, schedNameSelected) => acc || ((scheduleStateInit
+                            .schedules[schedNameSelected].id === id) && schedNameSelected), false);
+                    if (schedName) {
+                        store.dispatch(deleteSchedule(schedName));
+                    }
+                }
+            });
+        localStorage.setItem("coursePlanObservedSchedules", JSON.stringify(schedulesObserved));
+        window.setInterval(() => {
```
davish: Should make it a bit more clear that this is setting up a timer that keeps running repeatedly. Took me a bit to see that

frontend/plan/src/syncutils.js
```diff
@@ -0,0 +1,101 @@
...
+                    const schedule = scheduleState.schedules[scheduleName];
+                    if (!schedule.pushedToBackend) {
+                        if (schedule.isNew && !schedule.isNew.creationQueued && !("id" in schedule)) {
+                            store.dispatch(createScheduleOnBackend(scheduleName, schedule.meetings));
+                        } else {
+                            store.dispatch(updateScheduleOnBackend(scheduleName, schedule));
+                        }
+                    }
+                });
+        }, 2000);
```
davish: This `2000` should also be a constant

davish:
@dlike230 one more thing I realized is that if we display local schedules, than users won't ever know if a class in their schedule changes time, since they'll keep seeing their localstorage copy and the copy on the server (which will have the updated class time) will never be displayed. Was wondering what you thought about a possible solution to that? Something I thought of was to have the server-side schedule be switched to it's fully loaded. So users always see schedules, just at first they see the (maybe out-of-date) local schedule before it switches to the updated one from the server. but not sure if that's a bad UX

frontend/plan/src/components/search/SearchBar.js
```diff
@@ -37,6 +37,90 @@ function shouldSearch(filterData) {
...
+                        <p className="email-container">
+                            {" "}
+                            {username}
+                            {" "}
+                        </p>
+                        <div
+                            role="button"
+                            id="logout-button"
+                            onClick={() => {
+                                fetch("/accounts/logout").then(console.log);
```
eyingxuan: From my debugger, seems like backend is responding with a 302. Code as is does not redirect - I think we need to add a redirect: "follow".
[https://stackoverflow.com/questions/39735496/redirect-after-a-fetch-post-call](url)

frontend/plan/src/components/search/SearchBar.js
```diff
@@ -37,6 +37,90 @@ function shouldSearch(filterData) {
...
+            fontSize: "1rem!important",
+            paddingRight: "1rem",
+            paddingLeft: "1rem",
+        }}
+    >
+        Login
+    </a>
+);
+
+const UserSelector = ({ user: { first_name, last_name, username } }) => {
```
eyingxuan: The "click outside" logic of this component can be abstracted away with useOnClickOutside.js

frontend/plan/src/actions/index.js
eyingxuan: In general we have pretty bad error handling practices - in the next round of refactoring we should probably figure out how to catch all these exceptions in general.

frontend/plan/src/components/search/SearchBar.js
davish: can also fix by adding a trailing slash I'm 90% sure

frontend/plan/src/syncutils.js
```diff
@@ -0,0 +1,101 @@
...
+    deletionSuccessful,
+    fetchSchedulesAndInitializeCart,
+    incrementDeletionAttempts,
+    updateScheduleOnBackend
+} from "./actions";
+import fetch from "cross-fetch";
+import getCsrf from "./csrf";
+
+// Ensure that fetches only happen once ever 250 ms
+let lastFetched = 0;
```
eyingxuan: I think the design of this function could be improved - instead of returning null when theres a delay, consider having conditional fetch take in a function as a parameter and call the function only if the condition is met, where the function will call the fetch with the promise chain.

frontend/plan/src/syncutils.js
eyingxuan: On this note - just wanted to make sure that the backend stores/checks timestamps for updating schedules so if a request sent earlier reaches the backend later it won't overwrite the later request.

frontend/plan/src/syncutils.js
dlike230: This function is only being used for fetches though. The extra complexity required to do what you're describing isn't worth it in my opinion

frontend/plan/src/components/search/SearchBar.js
dlike230: I tried `redirect: "follow"`, the trailing slash, and some other things I found on StackOverflow, but none of them worked. For now, I'll just explicitly check for a 302 and handle that.

dlike230: Actually, it looks like that's not even necessary. It still gets picked up in the `.then` block since it's not an error code, and it ends up performing the logout correctly anyway.

frontend/plan/src/syncutils.js
dlike230: nvm you're right, that would be cleaner, now that I think about it more

frontend/plan/src/actions/index.js
```diff
@@ -351,23 +409,157 @@ export function fetchCourseDetails(courseId) {
...
-                        description: json.description,
-                        crosslistings: json.crosslistings,
-                    };
-                    const { sections } = json;
-                    dispatch(updateCourseInfo(sections, info));
-                },
+        fetch(buildSectionInfoSearchUrl(searchData))
+            .then(
+                response => response.json()
+                    .then(
```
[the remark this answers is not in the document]
dlike230: This isn't part of my PR

frontend/plan/src/reducers/schedule.js
dlike230: renamed for clarity

frontend/plan/src/reducers/schedule.js
```diff
@@ -89,6 +117,84 @@ const nextAvailable = (scheduleName, used) => {
...
+                    const cloudVersionUpdated = new Date(updated_at).getTime();
+                    if (name === "cart") {
+                        state.cartId = scheduleId;
+                        if (!state.cartUpdated || cloudVersionUpdated > state.cartUpdated) {
+                            newCart = sections;
+                        }
+                    } else if (state.schedules[name]) {
+                        newScheduleObject[name].id = scheduleId;
+                        newScheduleObject[name].isNew = false;
+                        if (!state.schedules[name].updated_at || state.schedules[name].updated_at < cloudVersionUpdated) {
```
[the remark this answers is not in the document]
dlike230: No good way to refactor this, so I just made the ordering of the operands more consistent

PR #109, opened 2019-11-02
Add Wharton TIA/GEBS requirements

ezwang:
- Add script to scrape Wharton TIA/GEBS requirements
- Add CircleCI badge to readme
- Make pipenv python requirement more generic

### DIFF
--- a/courses/requirements/wharton.py
+++ b/courses/requirements/wharton.py
@@ -1,11 +1,10 @@
 #!/usr/bin/env python3
 
-import re
 import json
+import re
 
 import requests
 import unidecode
-
 from bs4 import BeautifulSoup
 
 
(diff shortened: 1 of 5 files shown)


courses/requirements/wharton.py
```diff
@@ -55,15 +61,47 @@ def _clean_data(data):
...
+    reqs = []
+    soup = BeautifulSoup(resp.content.decode('utf-8'), features='html.parser')
+    for lst in soup.select('.wpb_wrapper ul.bullet-list'):
+        for item in lst.select('li'):
+            match = re.match(r'([A-Z]{0,4})\s?(\d{3})\s?:', item.text, re.I)
+            if match is not None:
+                res = match.groups()
+                reqs.append({
+                    'department': res[0],
+                    'course_id': int(res[1]),
```
davish: I don't know if we want to cast to an int here -- a course like `ECON-001` will get represented as `department: ECON, course_id: 1` in the system, even though we would want the full three-character code `001`.

ezwang: Fixed! Nice catch!

PR #110, opened 2019-11-03
Charleyc/req res fix

mureytasroc:


courses/util.py
```diff
@@ -133,6 +134,15 @@ def add_college_requirements(course, college_reqs):
         req.courses.add(course)
 
 
+def relocate_reqs_from_rests(rests, reqs):
+    if any(r.requirement_description == 'Humanities & Social Science Sector' for r in rests):
```
davish: It might be possible to generalize this a bit more, with a list and an outer for loop. In case this comes up that there's another restriction that we want to treat as a requirement?

courses/util.py
```diff
@@ -133,6 +134,15 @@ def add_college_requirements(course, college_reqs):
         req.courses.add(course)
 
 
+def relocate_reqs_from_rests(rests, reqs):
```
davish: Might want to consider spelling out `restrictions` at the end of this function name, since `rest` is a little unclear

courses/util.py
```diff
@@ -166,6 +176,8 @@ def upsert_course_from_opendata(info, semester):
     set_meetings(section, info['meetings'])
     add_associated_sections(section, info)
     add_restrictions(section, info['requirements'])
+    relocate_reqs_from_rests(info['requirements'],
+                             info['fulfills_college_requirements'].append('Humanities & Social Science Sector'))
```
davish: was the `.append('Humanities & Social Science Sector')` for testing? looks like this would always add Humanities to every class.

courses/tests.py
```diff
@@ -323,3 +323,41 @@ def test_course_no_good_sections(self):
...
+        self.rests[0].requirement_description = 'Natural Science & Math Sector'
+        relocate_reqs_from_rests(self.rests, self.reqs)
+        self.assertEqual(self.reqs, ['Natural Science & Math Sector'])
+
+    def test_hss(self):
+        self.rests[0].requirement_description = 'Humanities & Social Science Sector'
+        relocate_reqs_from_rests(self.rests, self.reqs)
+        self.assertEqual(self.reqs, ['Humanities & Social Science Sector'])
+
+    def test_mixed(self):
```
davish: Can you add a test where there's a restriction that's not one of these three? Just to make sure that it doesn't add them in by accident.

Also a test where there's already something in `self.reqs` list in addition to the ones being added. Just to cover other edge cases.

courses/util.py
mureytasroc: Wow sry abt that...

davish:
Sorry about the test failure -- i pushed to master without linting :/

PR #112, opened 2019-11-04
Removed 2.5 default stat

szarkesh:
would appreciate someone testing this themselves -- think i did it in a way that covers any edge cases, but can't be totally sure.


davish:
Also tested it out -- looks like it works well!

PR #116, opened 2019-11-08
Mobile version

szarkesh:
updated section design

image


frontend/plan/src/components/selector/Section.js
```diff
@@ -1,40 +1,77 @@
...
+                        <div>
+                            {section.activity}
+                        </div>
+                        <div>
+                            {meetings.length > 0 ? (
+                                <div>
+                                    <i className="fas fa-map-marker-alt grey-text" />
+                            &nbsp;
+                                    {((l) => {
+                                        const ret = new Set();
```
davish: A comment about what this code is doing (formatting the location info) would be helpful

davish:
One other thing I noticed was that there's still a tooltip with the prof's name when you mouse over the PCR instructor score. Could you also remove that before merging?

dlike230:
It looks like it's getting cut off horizontally:
image

PR #117, opened 2019-11-10
Charleyc/backendalert

mureytasroc:


alert/views.py
```diff
@@ -233,3 +236,14 @@ def send_json(style, msg):
...
+
+
+class RegistrationViewSet(viewsets.ModelViewSet):
+    serializer_class = RegistrationSerializer
+    http_method_names = ['get', 'post', 'put']
+    permission_classes = [IsAuthenticated]
+
+    def get_queryset(self):
+        queryset = Registration.objects.filter(user=self.request.user)
+        queryset = super().get_serializer_class().setup_eager_loading(queryset)
```
davish: Remove `setup_eager_loading()` since we now have AutoPrefetch, but if we want PCR annotations, we can do the `prefetch_related` directly in `get_queryset()`.

courses/models.py
```diff
@@ -343,3 +347,31 @@ class APIKey(models.Model):
     active = models.BooleanField(blank=True, default=True)
 
     privileges = models.ManyToManyField(APIPrivilege, related_name='key_set', blank=True)
+
+
+class UserData(models.Model):
+    """
+    A model that stores all user data from PCX users
+    """
+    user = models.OneToOneField(get_user_model(), on_delete=models.CASCADE, related_name='user')
```
davish: a related name of `data` might be better? so in a request, to get the user data you would say `request.user.data` rather than `request.user.user`

courses/models.py
```diff
@@ -343,3 +347,31 @@ class APIKey(models.Model):
...
+    email = models.EmailField(blank=True, null=True)
+
+    def validate_phone(value):
+        try:
+            phonenumbers.parse(value, 'US')
+        except phonenumbers.phonenumberutil.NumberParseException:
+            raise ValidationError('Enter a valid phone number.')
+
+    phone = models.CharField(blank=True, null=True, max_length=100, validators=[validate_phone])
```
davish: all fields should be together at the top of a model class for readability

courses/serializers.py
```diff
@@ -201,3 +202,38 @@ class Meta:
...
+
+class UserDataSerializer(serializers.ModelSerializer):
+    user = UserSerializer(read_only=True)
+
+    class Meta:
+        model = UserData
+        fields = [
+            'email',
+            'phone',
+            'user'
```
davish: Don't think it matters too much in the end, since it's a one-to-one relationship, but to me having the user data nested within the user seems a bit more natural

courses/serializers.py
```diff
@@ -201,3 +202,38 @@ class Meta:
...
+        ]
+
+
+class StatusUpdateSerializer(serializers.ModelSerializer):
+    section = serializers.ReadOnlyField(source='section__full_code')
+
+    class Meta:
+        model = StatusUpdate
+        fields = ['section', 'old_status', 'new_status', 'created_at', 'alert_sent', 'request_body']
```
davish: `request_body` doesn't neeed to be included in the serializer. it was mostly for debugging purposes when we were building the webhook

courses/tasks.py
```diff
@@ -117,3 +121,35 @@ def semester_sync(query='', semester=None):
     load_courses(query=query, semester=semester)
     load_requirements(school='SEAS', semester=semester)
     load_requirements(school='WH', semester=semester)
+
+
+def load_course_status_history(full_path):
```
davish: I think it makes more sense to move this function directly into the `management` comment -- the `tasks.py` file is generally for Celery tasks. Don't think we've talked about Celery before, but basically it's a way to spin recurring tasks off into a queue to be worked through one-by-one, which this function doesn't really fit.

courses/views.py
```diff
@@ -78,3 +79,26 @@ def get_queryset(self):
 class RequirementList(generics.ListAPIView, BaseCourseMixin):
     serializer_class = RequirementListSerializer
     queryset = Requirement.objects.all()
+
+
+class UserDetailView(generics.RetrieveAPIView, generics.UpdateAPIView):
```
davish: There should be some more restrictions, so that a user can only retrieve / update their own data, I think as this stands, any user can update the info of any other user

davish: Would be good to add tests about this to make sure that the API returns `403 Forbidden` in the proper contexts

courses/views.py
```diff
@@ -78,3 +79,26 @@ def get_queryset(self):
...
+
+
+class StatusUpdateView(generics.ListAPIView):
+    serializer_class = StatusUpdateSerializer
+    http_method_names = ['get']
+    lookup_field = 'full_code'
+
+    def get_queryset(self):
+        queryset = StatusUpdate.objects.filter(Q(section__course__full_code=self.kwargs['full_code']))
+        queryset = super().get_serializer_class().setup_eager_loading(queryset)
```
davish: ditto on `setup_eager_loading()`

alert/models.py
```diff
@@ -90,14 +114,16 @@ def resub_url(self):
         return '{}/s/{}'.format(settings.PCA_URL, url.short_id)
 
     def alert(self, forced=False, sent_by=''):
-        if forced or not self.notification_sent:
+        if forced or not (self.notification_sent or self.muted or self.deleted):
             text_result = Text(self).send_alert()
```
davish: The send_alert() method is going to need to be changed as well, to look at the `UserData` before looking at `Registration.phone` and `Registration.email`. I believe this is in `alert/alert.py`

alert/tests.py
```diff
@@ -532,21 +534,190 @@ def test_no_permission(self):
...
+        self.assertEqual(response.data['user']['username'], 'jacob')
+        self.assertEqual(response.data['user']['first_name'], '')
+        self.assertEqual(response.data['user']['last_name'], '')
+        response = self.client.get('/api/settings/')
+        self.assertEqual(200, response.status_code)
+        self.assertEqual(response.data['email'], None)
+        self.assertEqual(response.data['phone'], None)
+        self.assertEqual(response.data['user']['username'], 'jacob')
+        self.assertEqual(response.data['user']['first_name'], '')
+        self.assertEqual(response.data['user']['last_name'], '')
```
davish: Should also have tests to make sure that users cannot access the data of other users. Some examples I can think of off the top of my head:
- Can't read/update user contact info
- Can't see another user's registrations

Views should be returning `403 Forbidden` when data is inaccessible to them. You can return a django `HttpForbidden()` object

alert/models.py
```diff
@@ -90,14 +114,16 @@ def resub_url(self):
...
             email_result = Email(self).send_alert()
             logging.debug('NOTIFICATION SENT FOR ' + self.__str__())
             self.notification_sent = True
             self.notification_sent_at = timezone.now()
             self.notification_sent_by = sent_by
             self.save()
+            if not self.auto_mute:
```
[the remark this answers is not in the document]
mureytasroc: I guess the effect on the user is the same, but isn't resubscribing better for keeping accurate data?  Using the same object for multiple alerts would overwrite the notification_sent_at field.  Also, right now is_active assumes notification_sent is false.  So I just assumed it would fit better with how it is right now if auto_mute being False results in the user being automatically resubscribed... but muted isn't set here so I changed it to this:

image

Should I do this differently?

alert/serializers.py
```diff
@@ -0,0 +1,21 @@
...
+    section = SectionDetailSerializer(read_only=True)
+
+    class Meta:
+        model = Registration
+        fields = ['created_at', 'updated_at', 'section', 'deleted', 'muted', 'auto_mute']
+
+    @staticmethod
+    def setup_eager_loading(queryset):
+        queryset = queryset.prefetch_related(
+            Prefetch('section', Section.with_reviews.all()),
```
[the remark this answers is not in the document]
mureytasroc: I guess not... I changed it to use `MiniSectionSerializer` instead.

alert/tests.py
```diff
@@ -532,21 +534,190 @@ def test_no_permission(self):
...
+        self.assertEqual(response.data[0]['alert_sent'], True)
+        self.assertFalse(hasattr(response.data[0], 'request_body'))
+
+    def test_cis121_missing(self):
+        response = self.client.get('/statusupdate/CIS-121/')
+        self.assertEqual(200, response.status_code)
+        self.assertEqual(0, len(response.data))
+
 
-    def test_update_status(self):
```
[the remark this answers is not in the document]
mureytasroc: It's a duplicate... it's already in `CourseStatusUpdateTestCase`

alert/views.py
```diff
@@ -233,3 +236,14 @@ def send_json(style, msg):
         return do_register(course_code, email_address, phone, SOURCE_API, send_json, key)
     else:
         raise Http404('Only POST requests are accepted.')
+
+
+class RegistrationViewSet(viewsets.ModelViewSet):
```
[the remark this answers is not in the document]
mureytasroc: I guess if we are using `MiniSectionSerializer` this wouldn't be necessary, right?

courses/models.py
mureytasroc: The problem is validate_phone has to be defined above in Python... also lambda functions don't allow imperative programming at all so I couldn't do it that way.  Is there a better workaround or can I just add a comment that the phone field is below?

courses/models.py
mureytasroc: Good catch... I made it `profile` b/c I think that's convention (at least that's what I gleaned from the forums)

alert/views.py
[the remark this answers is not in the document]
davish: `MiniSectionSerializer` still gets data from the Section model like the instructors and the course title, so `AutoPrefetchMixin` will be good here

courses/models.py
davish: Good point! Keeping it as-is is good.

alert/models.py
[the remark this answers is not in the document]
davish: Like I said on Slack, you're right. We should keep it so that one Registration object will only ever trigger one notification. What you changed it to with the `self.muted` setting makes sense!

courses/views.py
mureytasroc: Can users access other users' data?  I overrode the get_object method in the UserView and it is a RetrieveAPIView/UpdateApiView so I don't think you can get other users' data...

alert/tests.py
mureytasroc: Does test_multiple_users_independent cover what you're getting at?  I'm not sure logistically how a user would access another user's data so I'm not really sure when to return a 403.  See above comment.

courses/views.py
mureytasroc: Does test_multiple_users_independent cover what you're getting at?

alert/models.py
```diff
@@ -78,8 +88,22 @@ def validate_phone(self):
 
     def save(self, *args, **kwargs):
         self.validate_phone()
+        if self.user is not None:
```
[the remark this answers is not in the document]
davish: Good point -- especially since it only does the checks if `self.user is None`, which shouldn't be the case on new alerts.

alert/alerts.py
```diff
@@ -47,11 +47,15 @@ def __init__(self, reg):
         super().__init__('alert/email_alert.html', reg)
 
     def send_alert(self):
-        if self.registration.email is None:
+        if self.registration.user.profile.email is not None:
```
davish: This might cause an error if `self.registration.user` is None -- would add change the line to `if self.registration.user is not None and self.registration.user.profile.email is not None` -- there may be a more pythonic way to do that, but it's tough since user is nullable, profile is nullable, and email is nullable.

alert/alerts.py
```diff
@@ -64,13 +68,17 @@ def __init__(self, reg):
         super().__init__('alert/text_alert.txt', reg)
 
     def send_alert(self):
-        if self.registration.phone is None:
+        if self.registration.user.profile.phone is not None:
```
davish: ditto here

alert/models.py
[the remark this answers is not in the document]
davish: You're right -- especially since this is only run on legacy alerts because of the check on `self.user`.

alert/serializers.py
```diff
@@ -0,0 +1,21 @@
...
+
+class RegistrationSerializer(serializers.ModelSerializer):
+    section = MiniSectionSerializer(read_only=True)
+
+    class Meta:
+        model = Registration
+        fields = ['created_at', 'updated_at', 'section', 'deleted', 'muted', 'auto_mute']
+
+    @staticmethod
+    def setup_eager_loading(queryset):
```
davish: Remember to delete `setup_eager_loading`

alert/tests.py
```diff
@@ -532,21 +534,341 @@ def test_no_permission(self):
...
+        self.assertEqual(response.data['last_name'], 'newname')
+        response = self.client.get('/api/settings/')
+        self.assertEqual(200, response.status_code)
+        self.assertEqual(response.data['profile']['email'], 'example@email.com')
+        self.assertEqual(response.data['profile']['phone'], '+13131234567')
+        self.assertEqual(response.data['username'], 'jacob')
+        self.assertEqual(response.data['first_name'], '')
+        self.assertEqual(response.data['last_name'], 'newname')
+
+    def test_update_settings_change_username(self):
```
davish: Usernames are set by Platform to be pennkeys, so users shouldn't be able to change it with an API request. I'd be worried someone changes their username to an unregistered user so that when that user logs in the first time, they're not able to access the site (or worse, they're logged in to an existing account)

alert/tests.py
```diff
@@ -532,21 +534,341 @@ def test_no_permission(self):
...
+        self.assertEqual(response.data['last_name'], '')
+        response = self.client.get('/api/settings/')
+        self.assertEqual(200, response.status_code)
+        self.assertEqual(response.data['profile']['email'], 'example@email.com')
+        self.assertEqual(response.data['profile']['phone'], '+13131234567')
+        self.assertEqual(response.data['username'], 'jacob')
+        self.assertEqual(response.data['first_name'], '')
+        self.assertEqual(response.data['last_name'], '')
+
+    def test_update_settings_change_first_name(self):
```
davish: Should add tests to make sure you can't change the settings of another user.

alert/tests.py
```diff
@@ -532,21 +534,341 @@ def test_no_permission(self):
...
+        self.assertEqual(200, response.status_code)
+        self.assertEqual(response.data['profile']['email'], 'example@email.com')
+        self.assertEqual(response.data['profile']['phone'], '+13131234567')
+        self.assertFalse('favorite_color' in response.data['profile'])
+        self.assertEqual(response.data['username'], 'jacob')
+        self.assertEqual(response.data['first_name'], '')
+        self.assertEqual(response.data['last_name'], '')
+        self.assertFalse('middle_name' in response.data)
+
+    def test_ignore_fields_email_update(self):
```
davish: why is this an `ignore_fields` test? looks like it's correctly updating emails

alert/tests.py
```diff
@@ -532,21 +534,341 @@ def test_no_permission(self):
...
+        self.assertEqual(response.data['first_name'], '')
+        self.assertEqual(response.data['last_name'], '')
+
+    def test_multiple_users_independent(self):
+        User.objects.create_user(username='murey',
+                                 password='top_secret')
+        client2 = APIClient()
+        client2.login(username='murey', password='top_secret')
+        response = self.client.put('/api/settings/',
+                                   json.dumps({'username': 'jacob',
```
davish: do you need to always include `username` in the `PUT` requests to an account?

alert/serializers.py
```diff
@@ -0,0 +1,21 @@
...
+from rest_framework import serializers
+
+from alert.models import Registration
+from courses.models import Section
+from courses.serializers import MiniSectionSerializer
+
+
+class RegistrationSerializer(serializers.ModelSerializer):
+    section = MiniSectionSerializer(read_only=True)
+
```
davish: should also contain the `user` as a `SlugRelatedField(slug_field='username')`.

alert/serializers.py
```diff
@@ -0,0 +1,21 @@
+from django.db.models import Prefetch
+from rest_framework import serializers
+
+from alert.models import Registration
+from courses.models import Section
+from courses.serializers import MiniSectionSerializer
+
+
+class RegistrationSerializer(serializers.ModelSerializer):
+    section = MiniSectionSerializer(read_only=True)
```
davish: Might be better to have `section` be a `SlugRelatedField(slug_field='full_code')`. `full_code` is kinda new, but it will always hold the full code of a course or a section, all caps with dashes. you can see where it's set in the `save()` method on the `Course` and `Section` models. Means we don't need to do any extra lookups.

alert/views.py
```diff
@@ -233,3 +237,12 @@ def send_json(style, msg):
         return do_register(course_code, email_address, phone, SOURCE_API, send_json, key)
     else:
         raise Http404('Only POST requests are accepted.')
+
+
+class RegistrationViewSet(AutoPrefetchViewSetMixin, viewsets.ModelViewSet):
```
davish: If we're planning on replacing the `submit` route with a `POST` request to this view set, then there should be a check to make sure that the `user` field corresponds to the current logged in user. Otherwise, throw a 403 error.

alert/tests.py
```diff
@@ -532,21 +534,341 @@ def test_no_permission(self):
...
+        self.assertEqual(response.data['username'], 'murey')
+        self.assertEqual(response.data['first_name'], '')
+        self.assertEqual(response.data['last_name'], '')
+        response = client2.get('/api/settings/')
+        self.assertEqual(200, response.status_code)
+        self.assertEqual(response.data['profile']['email'], 'example2@email.com')
+        self.assertEqual(response.data['profile']['phone'], '+12121234567')
+        self.assertEqual(response.data['username'], 'murey')
+        self.assertEqual(response.data['first_name'], '')
+        self.assertEqual(response.data['last_name'], '')
```
davish: Should add another `TestCase` for Registrations. Some ideas for tests that would be good to have:
- Test to make sure we only return registrations for the current semester
- Test for successfully submitting registrations for a given user
- Test for trying to submit a registration for another user

courses/admin.py
```diff
@@ -71,6 +92,13 @@ class StatusUpdateAdmin(admin.ModelAdmin):
     ]
 
 
+"""
```
davish: Reminder to delete commented-out code, looks like this is outdated :)

courses/models.py
```diff
@@ -379,3 +386,39 @@ class APIKey(models.Model):
...
+            try:
+                phone_number = phonenumbers.parse(self.phone, 'US')
+                self.phone = phonenumbers.format_number(phone_number, phonenumbers.PhoneNumberFormat.E164)
+            except phonenumbers.phonenumberutil.NumberParseException:
+                raise ValidationError('Invalid phone number (this should have been caught already)')
+        super(UserProfile, self).save(*args, **kwargs)
+
+
+@receiver(post_save, sender=User)
+def create_or_update_user_profile(sender, instance, created, **kwargs):
```
davish: I like the hook here! is there a test to make sure this hook is working correctly?

courses/serializers.py
```diff
@@ -149,3 +150,47 @@ class Meta:
...
+            'email',
+            'phone'
+        ]
+
+
+class UserSerializer(serializers.ModelSerializer):
+    profile = UserProfileSerializer(read_only=False)
+
+    def update(self, instance, validated_data):
+        prof, _ = UserProfile.objects.get_or_create(user=instance)
```
davish: I could be completely wrong on this but I'm not sure this is 100% necessary? I think since this isn't `many=True`, Django REST Framework can figure out how to update the object.

courses/serializers.py
```diff
@@ -149,3 +150,47 @@ class Meta:
...
+        prof_data = validated_data.pop('profile')
+        for key in ['username', 'first_name', 'last_name']:
```
davish: Wrote this above, but not sure we want to allow updates to these fields since they're coming from Platform (who in turn is getting them from Penn)

courses/serializers.py
```diff
@@ -149,3 +150,47 @@ class Meta:
...
+        fields = [
+            'username',
+            'first_name',
+            'last_name',
+            'profile'
+        ]
+
+
+class StatusUpdateSerializer(serializers.ModelSerializer):
+    section = serializers.ReadOnlyField(source='section__full_code')
```
davish: Just a reminder to check and make sure we're not issuing one DB query per StatusUpdate, since we're accessing a related field here. You can check this in Django Debug Toolbar, slack me if u can't find it

courses/views.py
```diff
@@ -78,3 +80,23 @@ def get_queryset(self):
...
     serializer_class = RequirementListSerializer
     queryset = Requirement.objects.all()
+
+
+class UserView(generics.RetrieveAPIView, generics.UpdateAPIView):
+    serializer_class = UserSerializer
+    permission_classes = [IsAuthenticated]
+
+    def get_queryset(self):
+        return get_user_model().objects.all().prefetch_related('profile')
```
davish: Can we do `filter(pk=self.request.user.pk)` instead of `all()`? Just to be sure we're not exposing access to all users

courses/serializers.py
mureytasroc: I tried it and it said you can't update nested models... is there any way to get around that besides overriding update?  I read on stackexchange that you have to override but SE django posts are often out of date...

alert/tests.py
mureytasroc: Does `test_multiple_users_independent` cover that?

alert/tests.py
mureytasroc: The bulk of it is the request:

```
response = self.client.put('/api/settings/',
                                   json.dumps({'profile': {'email': 'example2@email.com'}}),
                                   content_type='application/json')
```

Which tests that you can update fields without including other fields and you won't have any issues or unwanted updates.

alert/tests.py
mureytasroc: Fixed that.

courses/views.py
mureytasroc: Done

courses/serializers.py
davish: in that case then yeah this works

PR #118, opened 2019-11-10
Update Dependencies

ArmaanT:
This PR updates DLA to a version that can handle Penn IDs. All the rows in the production database should be deleted prior to merging.


ArmaanT:
The only db that needs to be dropped is the one in production, so we don't need to worry about breaking developer's local databases. However, I did just realize that #120 updated DLA as well and was merged into master, so this PR isn't needed any longer.

PR #119, opened 2019-11-10
Adding CU Tag for courses

eyingxuan:
Adding CU tag for each section


eyingxuan:
Screen Shot 2019-11-17 at 1 15 45 PM
This PR adds CU information to each section

PR #123, opened 2019-11-15
IntegrityError returns more informative error message

mureytasroc:


mureytasroc:
IntegrityError more informative error message

PR #125, opened 2019-11-17
Type Filter

eyingxuan:
Added filter for lab, recitation, studio or seminar


eyingxuan:
Screen Shot 2019-11-21 at 3 29 15 PM

eyingxuan:
Filters should work now and passes all backend tests!

PR #132, opened 2019-12-21
Update reviews by semester

davish:
This PR adds utility functions and manage.py support for updating PCR ratings for a single semester, thanks to a new PCR endpoint!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,4 @@ PennCourses/settings/static/
 *.sqlite3
 *-test.json
 celerybeat-schedule
-pcr-backup
+pcr-backup*
--- a/review/management/commands/pcrbackup.py
+++ b/review/management/commands/pcrbackup.py
@@ -2,7 +2,7 @@ import os
 
 from django.core.management.base import BaseCommand
 
-from review.import_reviews import save_all_reviews, save_reviews_for_department
+from review.import_reviews import save_all_reviews, save_reviews_for_department, save_reviews_for_semester
 
 
 class Command(BaseCommand):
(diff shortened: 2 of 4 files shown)


review/management/commands/pcrbackup.py
```diff
@@ -19,10 +20,14 @@ def handle(self, *args, **kwargs):
         else:
             answer = input('Output directory already exists. Overwrite existing contents? (y/N): ')
             if answer.lower() != 'y':
```
mureytasroc: This isn't something new so it's prob fine but how come you are allowing 'y' and 'Y'?  Isn't it usually convention to require y to be lowercase and N to be uppercase?

.gitignore
```diff
@@ -14,4 +14,4 @@ PennCourses/settings/static/
 *.sqlite3
 *-test.json
 celerybeat-schedule
-pcr-backup
+pcr-backup*
```
mureytasroc: Why is the star necessary?

review/management/commands/pcrbackup.py
davish: I could totally be wrong on this, but I think the `y/N` convention is just conveying that `N` is the default value if you press enter: it's not actually enforcing that `y` needs to be lowercase and `N` needs to be uppercase

.gitignore
davish: Mainly so we can have multiple backup directories. On my machine, `pcr-backup` contains all PCR data up until 2018A -- I made a new directory called `pcr-backup-semester` to test this feature out, wanted to make sure it was also in the gitignore

review/management/commands/pcrbackup.py
mureytasroc: Oh ok that makes sense

.gitignore
mureytasroc: Makes sense

PR #136, opened 2019-12-24
Fix Search Result Scroll [ch280]

davish:
This PR makes the search result pane maintain its scroll position through clicks back and forth to an individual class!

The problem was the search result list was getting mounted and un-mounted, so if the DOM element wasn't there, its scroll position wasn't being maintained by the browser. It's now maintained by state when the component gets un-mounted.


frontend/plan/src/components/selector/CourseList.js
```diff
@@ -31,37 +31,76 @@ const courseSort = (courses, sortMode) => {
...
-            >
-                <div className="header" style={{ overflow: "hidden", width: "60%" }}>
+export default class CourseList extends React.Component {
+    constructor(props) {
+        super(props);
+        this.listRef = React.createRef();
+    }
+
+
+    componentDidMount() {
```
eyingxuan: We only want to use scrollTop from selector state only if it comes from the "Back" button (clearCourse redux action I think) right? Or else you would be skipping courses for a totally new query.

frontend/plan/src/components/selector/CourseList.js
```diff
@@ -31,37 +31,76 @@ const courseSort = (courses, sortMode) => {
...
+    // If we have a snapshot value, we've just added new items.
+    // Adjust scroll so these new items don't push the old ones out of view.
+    // (snapshot here is the value returned from getSnapshotBeforeUpdate)
+        const { scrollPos } = this.props;
+        const list = this.listRef.current;
+        list.scrollTop = scrollPos;
+    }
+
+    componentWillUnmount() {
```
eyingxuan: Just a comment that doesn't require a rewrite - componentWillUnmount can be emulated with hooks by placing this code in a function in the return statement of useEffect.

frontend/plan/src/components/selector/CourseList.js
davish: great point -- i'll look into fixing this.

frontend/plan/src/components/selector/CourseList.js
sigmachirality: moved back to functional component using the useEffect cleanup functionality!

frontend/plan/src/components/selector/CourseList.js
sigmachirality: added useEffect which sets scrollTop to 0 when course changes, not sure if that works tho

sigmachirality: it no work :(

sigmachirality: moved to redux state

frontend/plan/src/components/selector/CourseList.js
```diff
@@ -31,22 +31,41 @@ const courseSort = (courses, sortMode) => {
...
+    courses, getCourse, sortMode, scrollPos, setScrollPos,
+}) => {
+    const listRef = useRef(null);
+    useEffect(() => {
+        // If we have a snapshot value, we've just added new items.
+        // Adjust scroll so these new items don't push the old ones out of view.
+        // (snapshot here is the value returned from getSnapshotBeforeUpdate)
+        listRef.current.scrollTop = scrollPos;
+        // Return cleanup function
+        return () => setScrollPos(listRef.current.scrollTop);
```
eyingxuan: This seems to be causing an infinite rerender loop since useEffect has no deps and the cleanup function gets called every rerender

sigmachirality: updated deps

PR #138, opened 2019-12-24
Automatically fetch course details if only one search result

davish:
[ch200]


eyingxuan:
LGTM - can be merged when ready!

PR #140, opened 2019-12-28
Hotfix/sched back create section

mureytasroc:


mureytasroc:
Improvements in schedules-backend (you can't create a section anymore, tests were added for that, tests were added to catch previous error on sections with multiple instructors, tests were added to better test equality between schedule object and serialized json).  The create_mock_data_with_reviews function was added for better testing in the future (and was utilized in better testing of schedules backend).  Also, Davis fixed a review averaging issue in the SQL querying (causing duplicates in Section.with_reviews and incorrect averaging of reviews in sections with multiple instructors).

PR #141, opened 2020-01-02
Refactor CSS to CSS Modules

sigmachirality:
Thinking about ways to do sticky footers in CSS still keeps ~~me~~ **Armaan** up at night

- [ ] Move all inline-styles to modules
- [ ] Move all global styles to modules
- [ ] Remove as many !important tags as possible
- [ ] Refactor/restructure styles

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,9 @@
+# Development Environment
 .env
 .vscode
 .idea/
 .DS_Store
+env.sh
 *.pyc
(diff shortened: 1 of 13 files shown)


sigmachirality:
@davish has suggested switching to modules instead of using styled-components.

sigmachirality:
@eyingxuan I'll reopen this if I get cuffed.

PR #142, opened 2020-01-06
Show Pre-reqs to User

sigmachirality:
Screenshot from 2020-02-15 05-45-59


frontend/plan/src/components/selector/CourseDetails.js
```diff
@@ -43,6 +44,16 @@ export default function CourseDetails({ course, getCourse }) {
                     </li>
                 ) : null
             }
+            {prerequisites.length ? (
+                <li>
+                    <span className="icon is-small">
+                        <i className="fas fa-random" />
+                    </span>
+                    &nbsp; Prerequisites: &nbsp;
+                    <TagList elements={[prerequisites.replace(/-/g, " ")]} limit={2} />
```
davish: Will there ever be more than one tag?

sigmachirality: I've actually written some code to extract course codes out of the prereq text because there were cases where the prereq text was really, really long and overflowed out of the tag. I think for prereqs, we should always show a show more button that shows the original text if the text more than ~1.5 times the length of the total course code lengths, though I'd like design to comment on that first.

sigmachirality:
@davish the upsert course info function in courses/util.py seems to already handle prerequisites, see https://github.com/pennlabs/penn-courses/blob/master/courses/util.py#L170

frontend/plan/src/components/selector/CourseDetails.js
```diff
@@ -5,11 +5,22 @@ import TagList from "./TagList";
 import Badge from "../Badge";
 
 const getReqCode = (school, name) => `${{ SAS: "C", SEAS: "E", WH: "W" }[school]}: ${name}`;
+const getPrereqCodes = (text) => {
+    if (typeof text !== "string") return [];
+    const captured = text.match(/(^|\W)[a-zA-Z]{3}[a-zA-Z]?(-|\s)[0-9]{3}($|(?=\W))/gm);
+    return captured ? captured.map(prereq => prereq.replace(/\W/g, " ").trim()) : [];
```
davish: Might be better to show the prereq string by default if the regex doesn't match, because it can still have useful info to students.

sigmachirality: Talked to Jess, and we're going to display the raw text but annotate the course codes as links

PR #143, opened 2020-01-11
Install options from pypi

ArmaanT:
This PR removes `options` from this repo and installs it through pypi.


setup.cfg
```diff
@@ -1,12 +1,12 @@
 [flake8]
 max-line-length = 120
-exclude = .venv, migrations, settings, manage.py, shortener, options, frontend
+exclude = .venv, migrations, settings, manage.py, shortener, frontend
```
davish: I think we want to remove `shortener` from here too

ArmaanT: Fixed

PR #144, opened 2020-01-13
Variable backend URL

davish:


frontend/plan/src/constants.js
```diff
@@ -0,0 +1 @@
+export const BACKEND_URL = process.env.REACT_APP_BACKEND_URL;
```
ArmaanT: Can this be `BACKEND_DOMAIN`? Just since we're following that notation elsewhere?

PennCourses/settings/prod.py
```diff
@@ -24,6 +26,24 @@
 # Honor the 'X-Forwarded-Proto' header for request.is_secure()
 SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
 
+# Share cookie with frontend
+SESSION_COOKIE_DOMAIN = '.pennlabs.org'
```
ArmaanT: We don't need to do any cookie sharing for this or CSRF

frontend/plan/src/constants.js
davish: I don't think DOMAIN is accurate in this case; for example, in prod, the value is `https://api.penncourses.org/plan`.

PennCourses/settings/prod.py
davish: This is in anticipation of the accounts merge-in where we'll have a session cookie

ArmaanT: Right, but the cookie doesn't need to be shared. The idea is the frontend (in the user's browser) makes requests to api.penncourses.org and django will get access to the cookie directly. So the cookie doesn't need to be manually shared with those domains.

frontend/plan/src/constants.js
ArmaanT: Oh I see, I didn't realize you were scoping the the `plan` path. That makes sense.

PR #147, opened 2020-01-19
Schedules backend semester inferring, improvements

mureytasroc:
request data finding, semester inferring, Schedule model validation removed

### DIFF
--- a/plan/views.py
+++ b/plan/views.py
@@ -1,3 +1,5 @@
+import json
+
 from django.core.exceptions import ObjectDoesNotExist
 from django.db import IntegrityError
 from django.db.models import Prefetch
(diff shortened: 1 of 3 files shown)


plan/views.py
```diff
@@ -54,35 +56,47 @@ def get_sections(data):
     return sections
 
 
+def check_semester(data, sections):
```
davish: might be nice to also return the semester it sets data's semester to, in case we use this function for something different going forward

mureytasroc: The problem is 
```
semester_res = self.check_semester(data, sections)
        if semester_res is not None:
            return semester_res
```
If the function were not partial, then this code wouldn't work...

PR #149, opened 2020-01-22
Shaya changes

szarkesh:
a) adding course to cart now scrolls to the part of the cart and highlights the last added course
b) alternate view that expands search and hides schedule
c) changed design for course conflicts (now a more subtle icon rather than graying out the course in the cart)


szarkesh:
made the button look better / stick out like a sore thumb less

PR #151, opened 2020-01-24
Live syncing in accounts frontend

dlike230:
You can open in multiple tabs now and have schedules sync between the two tabs (there's a delay though)

### DIFF
--- a/frontend/plan/src/syncutils.js
+++ b/frontend/plan/src/syncutils.js
@@ -44,7 +44,9 @@ const initiateSync = (store) => {
     let firstSync = !localStorage.getItem("usesBackendSync");
     localStorage.setItem("usesBackendSync", "true");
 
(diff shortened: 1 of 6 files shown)


frontend/plan/src/syncutils.js
```diff
@@ -61,67 +107,28 @@ const buildSyncLoop = (store) => {
...
-                                .schedules[schedNameSelected].id === id) && schedNameSelected),
-                            false);
-                        if (schedName) {
-                            store.dispatch(deleteSchedule(schedName));
-                        }
-                    }
-                });
-            localStorage.setItem("coursePlanObservedSchedules", JSON.stringify(schedulesObserved));
+    const startSyncLoop = async () => {
+        while (!syncTerminated[0]) {
```
davish: why is this an array instead of just a variable? do things get pushed to it?

PR #152, opened 2020-01-24
Warning on close when unsaved changes

dlike230:
The user is warned when they have changes that aren't pushed to the backend and they try to close the page

### DIFF
--- a/frontend/plan/src/syncutils.js
+++ b/frontend/plan/src/syncutils.js
@@ -64,6 +64,17 @@ const buildSyncLoop = (store) => {
 };
 
 /**
+ * Returns whether all schedules have been pushed to the backend
+ */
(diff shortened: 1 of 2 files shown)


frontend/plan/src/syncutils.js
```diff
@@ -119,6 +130,13 @@ const initiateSync = (store) => {
             intervalRecord.push(window.setInterval(syncLoop, SYNC_INTERVAL));
         }));
 
+    window.addEventListener("beforeunload", (e) => {
+        if (!allPushed(store.getState().schedule)) {
+            e.preventDefault();
+            e.returnValue = "";
```
davish: Do we want to put a message here? Not exactly sure how this event listener works

dlike230:
Now redundant

PR #156, opened 2020-01-27
fixed a search bug that reappeared

szarkesh:


frontend/plan/src/components/selector/Selector.js
```diff
@@ -72,7 +72,7 @@ function Selector(props) {
         if (courses.length === 1) {
             getCourse(courses[0].id);
         }
-    }, [courses, getCourse, isSearchingCourseInfo]);
+    }, [isSearchingCourseInfo]); // eslint-disable-line
```
eyingxuan: I haven't had time to look at the entire thing, but it is generally bad practice/an anti-pattern to exclude variables in the useEffect from the dependency array. Might have to restructure code to avoid this

dlike230: ^ yeah I agree

eyingxuan: Screen Shot 2020-01-27 at 11 28 03 AM

I tried something like this ^^ should work

eyingxuan: Screen Shot 2020-01-27 at 2 16 12 PM
Something like this should also work, then we'd remove the useEffect. 

My qualms with removing course from the dependency array is that course essentially becomes a stale state in the useEffect, and is dependent on the invariant that we need that function to be called only if we change the isSearchingCourseInfo boolean, which could be subjected to change.

eyingxuan: We should probably also dispatch updateSearch first. You'd still have to dispatch updateSearch because that sets the searchInfoLoading flag to false

PR #168, opened 2020-02-16
Move backend code to backend directory

davish:
Better directory organization, considering we have both frontend and backend code in the same repo. Matches what penn-clubs does.


backend/courses/tests.py
```diff
@@ -459,7 +459,8 @@ def test_mixed_other_already_populated(self):
 class ParseOpendataResponseTestCase(TestCase):
     def test_parse_response(self):
         upsert_course_from_opendata(
-            json.load(open("courses/test-opendata.json", "r"))["result_data"][0], TEST_SEMESTER
+            json.load(open("backend/courses/test-opendata.json", "r"))["result_data"][0],
```
ArmaanT: You should be able to make this relative by only specifying the file name and not the path since they're in the same directory. Platform does something similar in our org tests somewhere.

ArmaanT:
Also not related to this PR (should probably go in the one to move the frontend directories) but you can use the react-tools orb that platform recently created to help simplify the pcp/pca frontend linting config

backend/courses/tests.py
davish: The right way to do it is to use the OS path things and specify an absolute path... this is being weird since I'm still running tests from the project root with vscode but it shouldn't be tied to my workflow

PR #169, opened 2020-02-19
API Documentation

davish:
[ch1690]

Move API documentation from Notion into docstrings so that it's easier to access and also gets pulled in when Django REST Framework auto-generates the OpenAPI specification for the project.

- [x] PCP Search routes
- [x] PCP Schedules routes
- [x] PCA API routes


backend/Pipfile
```diff
@@ -18,7 +18,6 @@ django-debug-toolbar = "*"
 
 [packages]
 django = "*"
-psycopg2 = "*"
```
davish, line 21: Just an FYI taht this change shouldn't be committed -- pyscopg2 is necessary for production deployments

mureytasroc: Yeah for sure; I can't run the server when that is included (can't get it to work on mac with pycharm) so I was planning to remove that for development and then add it back before merging.  Is there a better way to do it?

davish:
Looks good to yeet! Fantastic work -- really love how thorough this is! I hope some of the stuff in `docs_settings.py` makes it upstream into a Django REST Framework release someday :D

PR #181, opened 2020-02-28
Lint rules/fixes and dotfiles on PCA

eyingxuan:
This PR introduces linting (ESLint), a jsconfig file (`https://code.visualstudio.com/docs/languages/jsconfig`) and an editorconfig file (`https://editorconfig.org/`) as well as fixing lint issues.


sigmachirality:
ew linting

PR #182, opened 2020-02-28
Various syncing fixes

dlike230:
Fixes to deletion, opening of pcp on a new device, repeated attempted creations/deletions, and some other misc. stuff

Also fixed logout

### DIFF
--- a/frontend/plan/src/syncutils.js
+++ b/frontend/plan/src/syncutils.js
@@ -53,7 +53,7 @@ const initiateSync = async (store) => {
                 handleSemester(options.SEMESTER);
             })
             .catch(() => {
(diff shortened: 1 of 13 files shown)


frontend/plan/src/syncutils.js
```diff
@@ -53,7 +53,7 @@ const initiateSync = async (store) => {
                 handleSemester(options.SEMESTER);
             })
             .catch(() => {
-                handleSemester("2020C");
+                handleSemester("2020A");
```
davish: Can we pop up a modal in this case? or something saying that something went wrong instead of defaulting to a semester that might not be right?

PR #185, opened 2020-03-21
Pca timeline 2

szarkesh:
Added the timeline, it's still buggy

also probably have ∞ lint errors because my linter isn't working for some reason

also i couldn't get the backend call to work on my device so this currently uses https:// api, you'll get CORS if you don't download the CORS chrome extension

image


eyingxuan:
Re your question about using dev backend - you can add the following to the alert package.json file
Screen Shot 2020-03-21 at 10 10 37 AM

All fetch calls can thus be shortened to `fetch('/api/alert/...)` and the requests will be proxied to the local development backend (which you need to start with python manage.py runserver 8000). This should resolve the CORS errors.

eyingxuan:
Actually, since we will all be working with the backend soon, I'll probably open a PR that addresses this after Davis posts the data dump and I can verify that the local backend supports alert API calls.

frontend/alert/src/Timeline.js
```diff
@@ -0,0 +1,222 @@
...
+`
+
+const LeftRight = styled.div`
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: center;
+`
+
+function convertTime(timeString){
```
davish, line 113:
```suggestion
const convertTime = timeString => {
```

Could all of these use the new function syntax?

frontend/alert/src/Timeline.js
```diff
@@ -0,0 +1,222 @@
...
+    if(convertTime(data[i][0]["created_at"])[0]!==convertTime(data[i-1][0]["created_at"])[0]){
+      divs.push(<DateStyle offset={yoff[i]}>{convertTime(data[i][0]["created_at"])[0]}</DateStyle>);
+    }
+  }
+  console.log("date divs length is " + divs.length);
+  return divs;
+}
+
+function Timeline({courseCode, setTimeline}){
+  let [data, setData] = React.useState(null);
```
davish, line 139:
```suggestion
  let [data, setData] = useState(null);
```
And import `useState` directly at the top, to match `useEffect` below this.

frontend/alert/src/Timeline.js
```diff
@@ -0,0 +1,222 @@
...
+  let [segLengths, setSegLengths] = React.useState(null);
+  let [yOffsets, setYOffsets] = React.useState(null);
+  let [loaded, setLoaded] = React.useState(false);
+  let [displayedCode, setDisplayedCode] = React.useState(null);
+  useEffect(()=>{
+      if(courseCode==null){
+        return;
+      }
+      setLoaded(false);
+      fetch(`https://penncourseplan.com/api/alert/statusupdate/${courseCode}`).then(res=>res.json()).then(result=>{
```
davish, line 149: the `.then` chain should be split over multiple lines

davish: Also since Ying's change got merged in, can use the local API

frontend/alert/src/Timeline.js
```diff
@@ -0,0 +1,222 @@
...
+        console.log(result)
+        result.sort((a,b)=>(a.created_at > b.created_at) ? 1 : -1);
+        let simplifiedData = result.reduce((ans, item, index) => { // preprocessing hte data
```
davish, line 152: A comment describing the kind of preprocessing happening would be useful here

frontend/alert/src/Timeline.js
```diff
@@ -0,0 +1,222 @@
...
+          if(index==0){
+            return ans;
+          }
+          if(item["old_status"] == result[index-1]["old_status"]){
+            return ans;
+          }
+          if(item["old_status"] == "C" && item["new_status"] == "O"){
+            ans.push([item, "opened"]);
```
davish, line 160:
```suggestion
            return ans.concat([[item, "opened"]]);
```
So we're not mixing immutable and mutable calls -- with a function like reduce it's better to stick to immutable operations that return a new list.

frontend/alert/src/Timeline.js
```diff
@@ -0,0 +1,222 @@
...
+          if(item["old_status"] == "O" && item["new_status"] == "C"){
+            ans.push([item, "closed"]);
+          }
+          return ans;
+        }, []).reverse()
+        setData(simplifiedData);
+        var i;
+        let segmentLengths = [];
+        for (i = 1; i < simplifiedData.length; i++){
+          segmentLengths.push(Math.round(20+5*Math.pow(1 + absoluteTime(result[i]["created_at"]) - absoluteTime(result[i-1]["created_at"]), 0.2)));
```
davish, line 171: pulling this calculation out into a function would be good for readability.

frontend/alert/src/Timeline.js
```diff
@@ -0,0 +1,222 @@
...
+  return <TimeStyle offset={offset}>{convertedTime[1]}</TimeStyle>
+}
+
+const FlexRow = styled.div`
+  display: flex;
+  flex-direction: row;
+  position: relative;
+`
+
+const LeftRight = styled.div`
```
davish, line 106: It's difficult to tell from the component name exactly what `LeftRight` is/does -- would be nice to rename it to something more descriptive.

frontend/alert/src/Timeline.js
```diff
@@ -0,0 +1,222 @@
...
+            <MyButton onClick={()=>setTimeline(null)}><i className="fas fa-times"></i></MyButton>
+            <Heading>Alert History</Heading>
+            <LeftRight>
+                <Subheading>{displayedCode}</Subheading>
+                {loaded && data[0]["new_status"] === "O" ? <OpenBadge>Open</OpenBadge> : <ClosedBadge>Closed</ClosedBadge>}
+            </LeftRight>
+            <ScrollContainer>
+              { data && yOffsets ?
+                      <FlexRow>
+                        <div style={{width:"100px"}}>
```
davish, line 198: Probably should factor this out into a styled component, and also use rem over px.

frontend/alert/src/Timeline.js
```diff
@@ -0,0 +1,222 @@
...
+const OpenBadge = styled(ClosedBadge)`
+  color:
+`
+
+const Segment = styled.div`
+  width: 2px;
+  height: ${props=>props.height + 5}px;
+  background: ${props=>props.type=="opened" ? "#78d381" : "#cbd5dd"};
+`
+const MyCircle = styled.div`
```
davish, line 79: Why `MyCircle` instead of just `Circle`?

andyjiang3:
Added alert history timeline!

Screen Shot 2021-03-03 at 9 39 42 PM

andyjiang3:
Changed timeline spacing to better match figma design

Screen Shot 2021-03-03 at 10 33 30 PM

frontend/alert/components/Timeline.tsx
```diff
@@ -0,0 +1,297 @@
...
+  let prevTime = courseData[index - 1][0]["created_at"]
+  let currTime = courseData[index][0]["created_at"]
+
+  let segLength = Math.min(Math.round(MIN_SEGMENT_LENGTH + (Math.abs(timeInHours(prevTime) - timeInHours(currTime)) * MULTIPLIER)), MAX_SEGMENT_LENGTH)
+
+  let circle = <> 
+    <InfoLabel>{!courseData[index][2] && convertTime(currTime)[0]}</InfoLabel>
+    <Circle open={courseData[index][1] == "opened"}><FontAwesomeIcon icon={faCircle}/></Circle>
+    <InfoLabel isTime={true}>{convertTime(currTime)[1]}</InfoLabel>
+  </>
```
dlike230, line 204: should be `const`s since they're never modified

frontend/alert/components/Timeline.tsx
```diff
@@ -0,0 +1,297 @@
...
+
+const getMonthDay = (timeString) => {
+  let d = new Date(timeString);
+  return d.toLocaleDateString('en-US', {month:"numeric", day: "numeric"})
+}
+
+const formatData = (courseData) => {
+  // convert the course status data into an array of 
+  // length 3 arrays with the status data object, the current status at each created_at data point
+  // and if the date is different
```
dlike230, line 171: Better to use a multiline comment for this

frontend/alert/components/Timeline.tsx
```diff
@@ -0,0 +1,297 @@
+import React, {useState, useEffect, useRef} from 'react';
```
dlike230, line 1: `useRef` is unused

frontend/alert/components/Timeline.tsx
```diff
@@ -0,0 +1,297 @@
...
+import styled from "styled-components";
+
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faTimes, faCircle } from '@fortawesome/free-solid-svg-icons'
+import { IndexKind, resolveModuleNameFromCache } from 'typescript';
```
dlike230, line 6: This import is unused

frontend/alert/components/Timeline.tsx
```diff
@@ -0,0 +1,297 @@
...
+const Timeline = ({
+  courseCode,
+  setTimeline}: TimelineProps) => {
+  
+  const [courseStatusData, setCourseStatusData] = useState([]);
+  const [loaded, setLoaded] = useState<boolean>(false);
+
+  useEffect( () => {
+
+      if (courseCode == null) {
```
dlike230, line 235: Instead of checking `courseCode == null` (you do this in a few other places too), it's better to check `!courseCode`, since this takes care of both null and undefined without requiring use of loose equality (undefined == null will return true, which can lead to ambiguous code sometimes).

frontend/alert/components/Timeline.tsx
```diff
@@ -0,0 +1,297 @@
...
+              {courseStatusData[0][1] == "opened" ? 
+              <StatusLabel open={true}>Open</StatusLabel> : <StatusLabel open={false}>Closed</StatusLabel>}
+            </CourseInfoContainer>
+
+            <TimelineScrollContainer>
+              <FlexRow>
+                <TimelineContainer>
+
+                    {courseStatusData.map((item, index) =>
+                      index != 0 && createTimelineEle(courseStatusData, index)
```
dlike230, line 281: When creating components in a loop, you should add a key prop

dlike230 (changes requested):
Overall, nice job! Just a few small things (other than my inline comments):

- The X button in the alert history pane should have `cursor: pointer` applied to it
- Please remove frontend/alert/public/index.html from git
- To me it feels like clicking outside of the alert history pane should close it, but I want to get some more opinions on that. Maybe @jess-tan and @ebchen can weigh in.
- I think we should have a scrollbar if the alert history gets too long, but let's see what @jess-tan and @ebchen think.

I'm also going to have @davish take a look in case he has any other feedback.

frontend/alert/components/Timeline.tsx
```diff
@@ -0,0 +1,296 @@
...
+
+    return formattedData
+}
+
+const timeInHours = (timeString) => {
+  let d = new Date(timeString);
+  return d.getHours();
+}
+
+const createTimelineEle = (courseData, index) => {
```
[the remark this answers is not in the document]
andyjiang3: Good idea!

andyjiang3:
> Overall, nice job! Just a few small things (other than my inline comments):
> 
> * The X button in the alert history pane should have `cursor: pointer` applied to it
> * Please remove frontend/alert/public/index.html from git
> * To me it feels like clicking outside of the alert history pane should close it, but I want to get some more opinions on that. Maybe @jess-tan and @ebchen can weigh in.
> * I think we should have a scrollbar if the alert history gets too long, but let's see what @jess-tan and @ebchen think.
> 
> I'm also going to have @davish take a look in case he has any other feedback.

Made changes, waiting on opinions for scrollbar and clicking outside pane!

frontend/alert/components/Timeline.tsx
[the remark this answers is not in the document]
jess-tan: I agree- clicking outside should close it, but keep the X button. A scroll bar (only visible on scroll) sounds like a good idea!

frontend/alert/components/AutoComplete.tsx
```diff
@@ -134,7 +143,14 @@ const Container = styled.div<{ inputHeight: string }>`
     height: ${(props) => props.inputHeight};
 `;
 
-const Suggestion = ({ onClick, courseCode, title, instructor, selected }) => {
+const Suggestion = ({
```
davish, line 146: Maybe add a `SuggestionProps` interface here?

frontend/alert/components/Timeline/TimelineEle.tsx
```diff
@@ -0,0 +1,129 @@
...
+        MAX_SEGMENT_LENGTH
+    );
+};
+
+interface TimelineEleProps {
+    courseStatusData: any;
+    index: number;
+}
+
+const TimelineEle = ({ courseStatusData, index }: TimelineEleProps) => {
```
davish, line 89: Might as well name this `TimelineElement`, much clearer and not many more characters

frontend/alert/components/Timeline/TimelineEle.tsx
```diff
@@ -0,0 +1,129 @@
...
+    const prevTime = courseStatusData[index - 1][0]["created_at"];
+    const currTime = courseStatusData[index][0]["created_at"];
+    const segLength = calcSegmentLength(prevTime, currTime);
+
+    return (
+        <>
+            {index - 1 == 0 ? (
```
davish, line 96: I'd maybe pull this out into a `const` variable for readability concerns. `index-1` is a bit unclear, giving it a name maybe with a comment would help ensure we know *why* it's `-1` when someone looks at this code down the line.

frontend/alert/components/Timeline/index.tsx
```diff
@@ -0,0 +1,287 @@
...
+                setLoaded(true);
+            })
+        );
+    }, [courseCode]);
+
+    return (
+        <AlertHistoryContainer
+            close={close}
+            ref={useScroll(
+                useOnClickOutside(
```
davish, line 223: the `useScroll` and `useOnClickOutside` hooks should get bound to `const` calues before the return statement and then have those values used inline instead of calling the hooks inline. Since this always returns the same hooks it doesn't expliclty break one of the [rules of hooks](https://reactjs.org/docs/hooks-rules.html), but in case this component gets changed going forward to do some conditional rendering, better to have the hooks specified in the top part of the function.

frontend/alert/components/Timeline/index.tsx
```diff
@@ -0,0 +1,287 @@
...
+                onClick={() => {
+                    setTimeline(null);
+                    courseCode = null;
+                }}
+            >
+                <FontAwesomeIcon icon={faTimes} />
+            </CloseButton>
+
+            {/* Only show if loaded */}
+            {loaded && courseStatusData && courseStatusData.length > 0 ? (
```
davish, line 244: this is a pretty long conditional, would think about pulling these out into their own component definitions. Not necessarily their own files, but just pull them above for readability.

frontend/shared-components/src/useOnClickOutside.js
```diff
@@ -6,27 +6,34 @@ import { useEffect, useRef } from "react";
  *
  * @param {function} onClickOutside The function called when an outside click is detected
  * @param {boolean} disabled Whether the hook should still listen for outside click
+ * @param {string} ignoreClassName  The className to ignore if it is clicked
  * @return {Object} A ref to be passed as a ref props to the component that uses useOnClickOutside
  */
 
-export function useOnClickOutside(onClickOutside, disabled) {
+export function useOnClickOutside(onClickOutside, disabled, ignoreEle = "") {
```
davish, line 13: same here on `ignoreEle` vs `ignoreElement`.

davish (approved):
Looks really good! mostly some minor readability nitpicks but awesome work and I'm excited to see this merged in :)

frontend/alert/components/Timeline/index.tsx
```diff
@@ -217,19 +233,7 @@ const Timeline = ({ courseCode, setTimeline }: TimelineProps) => {
...
-                    () => {
-                        setTimeline(null);
-                        courseCode = null;
-                    },
-                    close,
-                    "historyIcon"
-                )
-            )}
-        >
+        <AlertHistoryContainer close={close} ref={onScroll(onClickOutside)}>
```
davish, line 236: This is still a hook, the alias makes it a bit confusing, I'd move it up to `const scrollRef = useOnScroll(onClickOutside)` or something like that

andyjiang3: Ohh got it!

dlike230 (approved):
LGTM!

PR #187, opened 2020-03-22
Added accounts and login/logout for PCA

dlike230:
Also added a folder of `shared-components` for components that are shared between PCA and PCP. See updated README for setup details.


sigmachirality:
I'm worried that PCP's jank CSS will contaminate the PCA codebase. I like styled components, can we just make all the shared components styled components so the css scoping is contained to the components?

dlike230:
> I'm worried that PCP's jank CSS will contaminate the PCA codebase. I like styled components, can we just make all the shared components styled components so the css scoping is contained to the components?

Yeah good idea. The only thing we can't get rid of as easily is Bulma.

dlike230:
All functionality here is done now -- just need to do the refactoring.

frontend/plan/Dockerfile
```diff
@@ -9,6 +9,7 @@ COPY package*.json /app/
 RUN npm install
 
 COPY . /app/
+COPY ../shared-components/src /app/src/components
```
davish, line 12: Just tested a docker build locally and this doesn't work... will need to find a workaround. Probably eventually we should make this an NPM package that we publish as part of the build step?

dlike230: It doesn't need to be a published npm package, we can just put a github link in our package.json. Also we could also use git submodules as another alternative.

PR #188, opened 2020-03-28
Alert Management UI

eyingxuan:
This PR adds UI components for the alert management page. 

Progress:
Screen Shot 2020-03-29 at 2 47 25 PM

Pre-review
- [x] Check/fix Safari button issue


eyingxuan:
I'm pretty much done - just have to do some cleanup as well as fix very annoying visual differences in safari :/

eyingxuan:
UI work is done for alert management, proceeding with logic in another PR.

eyingxuan:
I've also sneaked in a CSRF file, so it gets into master and everyone can use the function for POST requests. This was directly copied from the PCP codebase

frontend/alert/src/components/ManageAlert.js
```diff
@@ -0,0 +1,348 @@
...
+const AlertSearch = () => (
+    <SearchFlex valign>
+        <Flex valign margin="0.2rem" className={styles.search}>
+            <Img src={Search} alt="" width="0.6rem" height="0.6rem" />
+            <SearchInput type="search" placeholder="Search" />
+        </Flex>
+    </SearchFlex>
+);
+
+const AlertAction = Object.freeze({ Resubscribe: 1, Cancel: 2 });
```
davish, line 208: Are these enums? if so, a comment would be useful to determine that

frontend/alert/src/components/ManageAlert.js
```diff
@@ -0,0 +1,348 @@
...
+const AlertRepeat = Object.freeze({ Inactive: 1, EOS: 2, Once: 3 });
+
+const ActionButton = ({ type }) => {
+    let img;
+    let primary;
+    let secondary;
+    let text;
+
+    switch (type) {
+        case 1:
```
davish, line 219: If these numbers are tied to the enums above, I feel like it might be better to use strings rather than numbers.

frontend/alert/src/components/ManageAlert.js
```diff
@@ -0,0 +1,348 @@
...
+        <ActionFlex valign halign background={secondary}>
+            <Flex valign margin="0.3rem" className={styles.actionbutton}>
+                <P size="0.6rem" color={primary} weight="600">{text}</P>
+                <Img src={img} width="0.6rem" height="0.6rem" alt="" />
+            </Flex>
+        </ActionFlex>
+    );
+};
+
+ActionButton.propTypes = { type: PropTypes.oneOf([1, 2]) };
```
davish, line 244: propTypes is just screeaaaming for typescript 😉

frontend/alert/src/components/ManageAlert.js
```diff
@@ -0,0 +1,348 @@
...
+
+
+const AlertItem = ({
```
davish, line 247: I think it would be great for these larger components to get some explanatory comments about what they are/what they do. Maybe even split them out into their own file, just to make the file sizes a bit more manageable?

frontend/alert/src/components/ManageAlert.module.css
```diff
@@ -0,0 +1,29 @@
+.actionbutton > * {
```
davish, line 1: are these rules that can't be done in StyledComponents? wondering the reasoning behind this bit of CSS modules.

eyingxuan: My bad - did not know you could have these in styled components! Fixed.

frontend/alert/src/components/ManageAlert.js
eyingxuan: Indeed :)

frontend/alert/src/components/ManageAlert.js
eyingxuan: Fixed!

frontend/alert/src/components/ManageAlert.js
eyingxuan: Fixed!

frontend/alert/src/components/ManageAlert.js
eyingxuan: Done!

PR #190, opened 2020-04-11
Autocomplete

dlike230:


frontend/alert/src/components/AutoComplete.js
```diff
@@ -0,0 +1,237 @@
...
+    background-color: white;
+    padding-top: 0;
+    max-height: 20rem;
+    overflow-y: scroll;
+    &::-webkit-scrollbar {
+        display: none;
+    }
+`;
+
+const SuggestionBox = styled.div`
```
davish, line 59: These two components have kind of confusing names, also along with `SuggestionsContainer` and a few others. Not sure if there's a way around this -- since styled components can make that funky.

frontend/alert/src/components/AutoComplete.js
```diff
@@ -0,0 +1,237 @@
...
+const AutoCompleteContainer = styled.div`
+    position: relative;
+    display: block;
+    margin-bottom: 1rem;
+    left: -50%;
+    height: ${props => props.inputHeight};
+`;
+
+const Suggestion = ({ courseCode, title, instructor }) => (
+    <SuggestionBox>
```
davish, line 125: On second thought, I don't think all of these need to be prefixed with `Suggestion`. If there's no other `Box`, `LeftCol`, `Title` or `Subtitle` components, i feel like we can drop the suffix for brevity's sake.

frontend/alert/src/components/AutoComplete.js
```diff
@@ -0,0 +1,237 @@
...
+    z-index: 1;
+`;
+
+const AutoCompleteInputBackground = styled(AutoCompleteInput)`
+    background: white;
+    z-index: 0;
+    color: grey;
+`;
+
+const AutoCompleteContainer = styled.div`
```
davish, line 116: This is more of a TODO for after an MVP but we should look to be generalizing divs like this into a `Div` component where we can set different things by props.

frontend/alert/src/components/AutoComplete.js
```diff
@@ -0,0 +1,237 @@
...
+
+    return (
+        <AutoCompleteContainer
+            inputHeight={inputRef && `${inputRef.getBoundingClientRect().height}px`}
+            ref={useOnClickOutside(() => setActive(false), !show)}
+        >
+            <AutoCompleteInput
+                autocomplete="off"
+                placeholder="Course"
+                ref={setInputRef}
```
davish, line 204: Might be able to simplify this with the `useRef` hook to simplify this a bit: https://reactjs.org/docs/hooks-reference.html#useref

dlike230: I had it as that initially, but it didn't work because it didn't force rerenders.

PR #194, opened 2020-04-17
Alert Management Logic

eyingxuan:


frontend/alert/src/components/ManageAlertLogic.js
```diff
@@ -0,0 +1,97 @@
...
+            });
+            sections = Array.from(sections);
+
+            const promises = [];
+            sections.forEach((section) => {
+                promises.push(getStatusPromise(section));
+            });
+
+            Promise.all(promises)
+                .then(vals => Promise.all(vals.map(val => val.json())))
```
davish, line 37: Could this request happen inside of the `AlertItem` component? Since it's tied directly to that component. Might make this code cleaner to read / shorter.

eyingxuan:
Branch is ready for review! Features added include 
- Fetching registrations data from backend
- Action buttons to resubscribe/cancel for individual alerts
- Batch select/actions to resubscribe/cancel/delete
- Filter alerts by section id

frontend/alert/src/App.js
```diff
@@ -3,12 +3,9 @@ import "./App.css";
 import styled from "styled-components";
 import PropTypes from "prop-types";
 import Logo from "./assets/PCA_logo.svg";
-
+import ManageAlertWrapper from "./components/managealert/index";
```
davish, line 6:
```suggestion
import ManageAlertWrapper from "./components/managealert";
```
index.js files are the default for the directory, so no need to add `index`! I'd also recommend renaming the directory to `ManageAlerts` to match casing for components.

frontend/alert/src/components/managealert/AlertItem.js
```diff
@@ -60,10 +60,21 @@ export const AlertItem = ({
     return (
         <>
             <GridItem column="1" row={rownum} border halign valign>
-                <input type="checkbox" />
+                <input type="checkbox" checked={checked} onChange={toggleAlert} />
             </GridItem>
             <GridItem column="2" row={rownum} border valign>
-                <P size="0.7rem">{date}</P>
+                {
+                    date
```
davish, line 67: what date does this represent? from the context I'd guess it has to do with the last time an alert was sent for this registration. Might be better in that case to rename this prop `alertLastSent` or something along that line.

frontend/alert/src/components/managealert/AlertItem.js
```diff
@@ -21,7 +21,7 @@ const StatusGridItem = styled(GridItem)`
 
 // Component for an alert entry (renders as a row in CSS grid)
 export const AlertItem = ({
-    date, course, status, repeat, actions, rownum,
+    date, course, status, repeat, actions, rownum, checked, toggleAlert, actionHandler,
```
davish, line 24: would rename `actionHandler` to something more specific -- what action is this for? When is it invoked?

frontend/alert/src/components/managealert/AlertSearch.js
```diff
@@ -25,11 +26,16 @@ const SearchBarFlex = styled(Flex)`
 
 // Component for search filter
 // in alert management
-export const AlertSearch = () => (
+export const AlertSearch = ({ value, onChange }) => (
     <SearchFlex valign>
         <SearchBarFlex valign margin="0.2rem">
             <Img src={Search} alt="" width="0.6rem" height="0.6rem" />
-            <SearchInput type="search" placeholder="Search" />
+            <SearchInput type="search" placeholder="Search" value={value} onChange={onChange} />
```
davish, line 33: This can be a TODO for another PR to use the `AutoComplete` component from @dlike230.

frontend/alert/src/components/managealert/index.js
```diff
@@ -0,0 +1,189 @@
...
+const fetchAlerts = () => (
+    fetch("/api/alert/api/registrations")
+        .then(res => res.json())
+);
+
+const processAlerts = (setAlerts) => {
+    const fetchPromise = () => (
+        fetchAlerts()
+            .then(res => (
+                res.map((section) => {
```
davish, line 16: `section` might be misleading here. Would rename to `registration`

frontend/alert/src/components/managealert/index.js
```diff
@@ -0,0 +1,189 @@
...
+                        }
+                    } else {
+                        repeat = AlertRepeat.Inactive;
+                    }
+
+                    return {
+                        id: section.id,
+                        original_created_at: section.original_created_at,
+                        section: section.section,
+                        date: datetime,
```
davish, line 49: would rename to `lastNotified`

frontend/alert/src/components/managealert/index.js
```diff
@@ -0,0 +1,189 @@
...
+                        } else {
+                            repeat = AlertRepeat.Once;
+                        }
+                    } else {
+                        repeat = AlertRepeat.Inactive;
+                    }
+
+                    return {
+                        id: section.id,
+                        original_created_at: section.original_created_at,
```
davish, line 47: this would be a good spot to transition from python snake_case to JS camelCase.
```suggestion
                        originalCreatedAt: section.original_created_at,
```

davish: I'm guessing this would require some find-and-replace in other places

frontend/alert/src/components/managealert/index.js
```diff
@@ -0,0 +1,189 @@
...
+            Accept: "application/json",
+            "Content-Type": "application/json",
+            "X-CSRFToken": getCsrf(),
+        },
+        body: JSON.stringify(body),
+    });
+};
+
+
+const actionHandler = callback => (id, actionenum) => {
```
davish, line 110: This is basically a mini bit of redux here -- a descriptive comment would go a ways towards illuminating that for someone adding more functionality down the line.

frontend/alert/src/components/managealert/index.js
```diff
@@ -0,0 +1,189 @@
...
+    useEffect(() => {
+        processAlerts(setAlerts);
+    }, []);
+
+    return (
+        <>
+            <ManageAlertHeader />
+            <ManageAlert
+                setFilter={setFilter}
+                numSel={numSelected}
```
davish, line 172: If this is just derived from `alertSel`, it might make sense to just have it computed in the `ManageAlert` component.

frontend/alert/src/App.js
eyingxuan: oh wow! i did not know they default to index.js - learned something new :)

eyingxuan: We actually get a linter error for having a directory name with uppercases - seems like it's due to the fact that some file systems are case insensitive, and that screws with the code base.

eyingxuan: jk, this seems to be there since the beginning before I even moved directories. It seems to have not come from ESLint, but my editor. I'll investigate

eyingxuan: Never mind - the directory was the issue, after resetting to head my projectile cache still used the upper case directory which is why my linter was complaining. I'm gonna use lower case directory names then.

frontend/alert/src/components/managealert/AlertSearch.js
eyingxuan: This might be the wrong place for the `AutoComplete` component. I think you might be thinking of `ManageAlertHeader` which is where users actually search for alerts. This search merely lets them filter their own registrations

PR #196, opened 2020-04-18
Adjustments to autocomplete

dlike230:
Improvements to performance and arrow key/selection functionality


frontend/alert/src/components/AutoComplete.js
```diff
@@ -206,14 +241,29 @@ const AutoComplete = () => {
...
+                        const newIndex = getCurrentIndex() + 1;
+                        const newSelectedSuggestion = Math.min(newIndex, suggestions.length - 1);
+                        handleSuggestionSelect(newSelectedSuggestion);
+                    } else if (e.keyCode === 38 && suggestions) {
+                        // select a suggestion when the down arrow key is pressed
+                        const newSelectedSuggestion = Math.max(getCurrentIndex() - 1, -1);
+                        handleSuggestionSelect(newSelectedSuggestion);
+                    } else {
+                        const newValue = e.target.value;
+                        setValue(newValue);
```
davish, line 260: looks like `value` isn't necessarily going to be the `value` inside of the input element (that we get from `e.target.value`) and will only match up in this else case. This is fine, but I'm pretty sure it's non-standard for handling forms in React -- an explanatory comment where the state is defined would go a long way here, and potentially also renaming `value` to be `currentSearch` or something might also help.

frontend/alert/src/components/AutoComplete.js
```diff
@@ -206,14 +241,29 @@ const AutoComplete = () => {
...
                 placeholder="Course"
                 ref={setInputRef}
-                value={value}
-                onChange={({ target: { value: newValue } }) => {
-                    setValue(newValue);
-                }}
-                onKeyDown={(e) => {
-                    // autocomplete with backdrop when the right arrow key is pressed
+                onKeyUp={(e) => {
                     if (e.keyCode === 39 && inputRef && suggestions && suggestions[0]) {
```
davish, line 245:
```suggestion
                    if (e.keyCode === RIGHT_ARROW && inputRef && suggestions && suggestions[0]) {
```

Magic numbers for `keyCode`s should be defined as constants

frontend/alert/src/components/AutoComplete.js
```diff
@@ -123,25 +123,44 @@ const Container = styled.div`
...
-        </DropdownItemLeftCol>
-        <IconContainer>
-            <FontAwesomeIcon icon={faHistory} color="#c4c4c4" />
-        </IconContainer>
-    </DropdownItemBox>
-);
+    onClick, courseCode, title, instructor, selected,
+}) => {
+    const ref = useRef();
+    useEffect(() => {
```
davish, line 129: Not sure what this useEffect is for -- a comment would be helpful

frontend/alert/src/components/AutoComplete.js
```diff
@@ -170,6 +189,7 @@ const AutoComplete = () => {
     const [inputRef, setInputRef] = useState(null);
     const [value, setValue] = useState("");
     const [suggestions, setSuggestions] = useState([]);
+    const [rawSuggestions, setRawSuggestions] = useState(null);
```
davish, line 192: what's the difference between `rawSuggestions` and `suggestions`? Probably a clarification in a comment would be good.

PR #199, opened 2020-04-22
Registration and profile autofill

davish:
[ch1824] [ch1982] [ch1827]

This PR does a few things:
- Factors out `AlertForm` into its own component
- Contact autofill and updating from user profile
- Registration action
- Display registration messages as toasts
- [x] Manage.py command to ensure profiles have a user's school emails by default. Make sure to run this before we launch!

Some minor modifications were made to `Toast` and `AutoComplete` to integrate them with the request logic.

### DIFF
--- a/frontend/alert/src/App.js
+++ b/frontend/alert/src/App.js
@@ -103,7 +103,7 @@ function App() {
             {page === "home" ? (
                 <Flex col grow={1}>
                     <Heading />
-                    <AlertForm user={user} />
+                    { user ? <AlertForm user={user} /> : null }
                 </Flex>
             ) : <ManageAlertWrapper />
             }
--- a/frontend/alert/src/App.js
+++ b/frontend/alert/src/App.js
@@ -95,9 +95,12 @@ function App() {
     const [page, setPage] = useState(window.location.hash === "#manage" ? "manage" : "home");
     const [messages, setMessages] = useState([]);
 
+    const MESSAGE_EXPIRATION_MILLIS = 8000;
     const removeMessage = k => setMessages(msgs => msgs.filter(m => m.key !== k));
     const addMessage = ({ message, status }) => {
-        setMessages(msgs => [{ message, status, key: genId() }].concat(msgs));
+        const id = genId();
+        setMessages(msgs => [{ message, status, key: id }].concat(msgs));
+        setTimeout(() => removeMessage(id), MESSAGE_EXPIRATION_MILLIS);
     };
 
     const setResponse = (res) => {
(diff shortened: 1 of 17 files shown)


frontend/alert/src/components/Toast.js
```diff
@@ -64,52 +66,53 @@ const RightItem = styled.div`
 
 export const ToastType = Object.freeze({ Success: 1, Warning: 2, Error: 3 });
 
-export const Toast = ({ type, course }) => {
+const Toast = ({
+    onClose, children, isSuccess, isWarning, isError, type,
```
eyingxuan, line 70: Wasn't sure what isSuccess, isWarning and isError is for if they have the same information as the type enum

eyingxuan, line 70: Wasn't sure what isSuccess, isWarning and isError is for if they have the same information as the type enum

frontend/alert/src/App.js
```diff
@@ -131,30 +87,24 @@ const Heading = () => (
...
 
 function App() {
-    const onSubmit = () => {};
     const [user, setUser] = useState(null);
-    const [page, setPage] = useState("home");
+    const [page, setPage] = useState(window.location.hash === "#manage" ? "manage" : "home");
+    const [messages, setMessages] = useState([]);
+
+    const removeMessage = k => setMessages(msgs => msgs.filter(m => m.key !== k));
+    const addMessage = ({ message, status }) => {
```
eyingxuan, line 99: Should we consider a setTimeout here to remove the message after x seconds?

frontend/alert/src/components/MessageList.js
```diff
@@ -0,0 +1,77 @@
+import React from "react";
+import PropTypes from "prop-types";
+import styled from "styled-components";
+
+import Toast, { ToastType } from "./Toast";
+
+const ToastContainer = styled.div`
+position: absolute;
```
eyingxuan, line 8: I think this needs some margin

frontend/alert/src/components/Toast.js
davish: Not used for anything atm... I'll take them out. The idea was that it would be nicer to type <Toast isSuccess /> if you were declaring a toast than it would be to do the enum, but it's also kind of confusing

frontend/alert/src/App.js
davish: I was thinking about that –– wonder what @jess-tan would think

frontend/alert/src/components/MessageList.js
davish: will add!

eyingxuan:
Sorry, I was curious what one of the buttons did 😆

PR #202, opened 2020-04-24
Force login & fix autofill

dlike230:
[ch1924]

Haven't actually fully tested this because my version of the backend has decided it doesn't want to let me log in anymore

### DIFF
--- a/frontend/alert/src/components/shared/accounts/AccountIndicator.js
+++ b/frontend/alert/src/components/shared/accounts/AccountIndicator.js
@@ -15,17 +15,20 @@ const AccountIndicator = ({
     onLeft, backgroundColor, nameLength,
 }) => {
     useEffect(() => {
+        if (user) {
+            return;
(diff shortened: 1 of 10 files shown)


frontend/alert/src/components/shared/accounts/AccountIndicator.js
```diff
@@ -18,11 +18,13 @@ const AccountIndicator = ({
...
                 if (!response.ok) {
+                    login(null);
                     return;
                 }
                 response.json()
                     .then(newUser => login(newUser));
-            });
+            })
+            .catch(() => login(null));
     }, [login]);
```
davish, line 28: Looks like this fetch is now getting called in a loop when i run the code –– not sure why since the dependencies for useEffect didn't change.

dlike230: fixed!

PR #203, opened 2020-04-24
Deploy PCA 3.0

davish:
[ch1784]
Merge this when we want to deploy the new version of PCA to penncoursealert.com!

### DIFF
--- a/k8s/values.yaml
+++ b/k8s/values.yaml
@@ -25,7 +25,7 @@ applications:
       hosts:
         - host: penncourseplan.com
           paths: ["/api", "/admin", "/accounts", "/assets"]
-        - host: beta.penncoursealert.com
+        - host: penncoursealert.com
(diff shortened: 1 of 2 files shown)


k8s/values.yaml
```diff
@@ -36,21 +36,21 @@ applications:
...
+  #   secret: penn-courses
+  #   ingress:
+  #     hosts:
+  #       - host: penncoursealert.com
+  #         paths: ["/"]
+  #   extraEnv:
+  #     - name: ROOT_URLCONF
+  #       value: PennCourses.urls.pca
+  #     - name: DJANGO_SETTINGS_MODULE
+  #       value: PennCourses.settings.production
```
ArmaanT, line 50: We can probably just delete these lines

davish: in case we need to switch back in a hurry, I wanted to keep them in.

PR #205, opened 2020-04-28
ISC Dump Import Script

davish:
[ch247] [ch2023]

This PR adds a `manage.py` command (`iscimport`) which loads in PCR data directly from the SQL dumps we get from ISC, along with enough unit tests to make @ArmaanT blush.

Before merging:
- [x] @pennlabs/platform to create an s3 bucket and send over credentials
- [x] brainstorm a new name
- [x] Support for course descriptions. This will probably require a decent amount of processing in Python, since the sql dump for descriptions stores a separate row per-paragraph.
Maybe in this PR:
- [x] Support for the big boii PCR_RATINGS file, which has review details regarding rading distributions and that go beyond just averages.


backend/courses/util.py
```diff
@@ -34,22 +34,28 @@ def separate_course_code(course_code):
     raise ValueError(f"Course code could not be parsed: {course_code}")
 
 
-def get_or_create_course(dept_code, course_id, semester):
+def get_or_create_course(dept_code, course_id, semester, report_created=False):
```
ArmaanT, line 37: I'm not really a fan of `report_created` I think we should stick to the django spec and always report it

backend/courses/util.py
```diff
@@ -34,22 +34,28 @@ def separate_course_code(course_code):
...
-    course, _ = Course.objects.get_or_create(department=dept, code=course_id, semester=semester)
+    if report_created:
+        return course, c
 
     return course
 
 
-def get_or_create_course_and_section(course_code, semester, section_manager=None):
+def get_or_create_course_and_section(
+    course_code, semester, section_manager=None, report_created=False
```
ArmaanT, line 48: Same comment here with `report_created`

backend/review/import_utils/import_to_db.py
```diff
@@ -0,0 +1,312 @@
...
+
+# Maps column name from SUMMARY sql tables to common slug.
+COLUMN_TO_SLUG = dict([(x[0], x[2]) for x in REVIEW_BIT_LABEL])
+# Maps "context" value from RATING table to common slug.
+CONTEXT_TO_SLUG = dict([(x[1], x[2]) for x in REVIEW_BIT_LABEL])
+
+
+def titleize(name):
+    """
+    Titleize a course name or instructor, taking into account exceptions such as II.
```
ArmaanT, line 63: Would it be possible to explicitly state the exceptions? Regex isn't super readable.

backend/review/import_utils/import_to_db.py
```diff
@@ -0,0 +1,312 @@
...
+
+    if course_dirty:
+        stat("courses_updated")
+        course.save()
+
+    return course, section
+
+
+def import_review(section, instructor, enrollment, responses, form_type, bits, stat):
+    # Assumption: that all review objects for the semesters in question were
```
ArmaanT, line 160: Does this apply to historical data or just the current semester?

backend/review/migrations/0005_auto_20200428_0003.py
```diff
@@ -0,0 +1,48 @@
+# Generated by Django 2.2.12 on 2020-04-28 04:03
```
ArmaanT, line 1: Can we squash this with the previous migration?

backend/courses/util.py
```diff
@@ -257,7 +263,7 @@ def create_mock_data_with_reviews(code, semester, number_of_instructors):
         section.instructors.add(instr)
```
ArmaanT, line 263: What do these create mock xyz functions do/what are they used for?

backend/courses/models.py
```diff
@@ -36,6 +38,7 @@ class Instructor(models.Model):
     updated_at = models.DateTimeField(auto_now=True)
 
     name = models.CharField(max_length=255, unique=True, db_index=True)
+    user = models.ForeignKey(User, null=True, blank=True, on_delete=models.SET_NULL)
```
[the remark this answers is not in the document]
davish: `User` here is actually an alias of `get_user_model()` (defined at the top of the file), which calls `settings.AUTH_USER_MODEL`

backend/review/tests.py
```diff
@@ -1,24 +1,238 @@
+from io import StringIO
```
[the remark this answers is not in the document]
davish: yeah, this is on my to-do list

backend/review/import_utils/import_to_db.py
davish: I copy/pasted this from PCR's code lmao but I'll try

backend/review/import_utils/import_to_db.py
davish: There is not `PCR_RATING_HIST` file... all rating details from 2002 onwards are in this one gb-large file 🙃

backend/courses/util.py
davish: creating review data for courses tests to make sure reviews get aggregated properly -- not related to this PR

backend/review/import_utils/import_to_db.py
davish: No difference between the two -- see the main Command class where we delete all reviews which have a matching semester field before importing.

backend/review/import_utils/import_to_db.py
davish: Maybe it's a good idea to write tests for this function directly too.

backend/courses/models.py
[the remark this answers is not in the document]
ArmaanT: Seems like the explicitly [recommend](https://docs.djangoproject.com/en/3.0/topics/auth/customizing/#django.contrib.auth.get_user_model) using `AUTH_USER_MODEL` over `get_user_model()`. Not entirely sure why though

backend/review/import_utils/import_to_db.py
ArmaanT: Is that good behavior though? I feel like it would make more sense to just import semesters you specify by default. I don't think it's possible for past data to change

backend/review/tests.py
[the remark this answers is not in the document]
ArmaanT: If that's set up as a story on clubhouse we can mark this as resolved

backend/courses/util.py
ArmaanT: That should probably get moved into a test file then, but it can probably just be done when moving the tests to a different folder

backend/review/import_utils/import_to_db.py
ArmaanT: Resolved offline.

davish: resolved off-line

backend/courses/migrations/0030_auto_20200503_1823.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 2.2.12 on 2020-05-03 22:23
```
ArmaanT, line 1: Can you squash this migration?

backend/review/import_utils/import_to_db.py
```diff
@@ -0,0 +1,356 @@
...
+            inst = Instructor.objects.get(user=user)
+            inst_created = False
+        elif Instructor.objects.filter(name=fullname).exists():
+            inst = Instructor.objects.get(name=fullname)
+            inst.user = user
+            inst.save()
+            inst_created = False
+        else:
+            inst = Instructor.objects.create(user=user, name=fullname)
+            inst_created = True
```
ArmaanT, line 114: Can this be simplified with a get_or_create?

backend/review/import_utils/parse_sql.py
```diff
@@ -0,0 +1,154 @@
...
+    """
+    Read in and parse a SQL dump, with each row as a Python dictionary.
+    tqdm shows a progress bar in the shell, and the process_row callback
+    is used to collect the resulting Python objects from the parser.
+
+    The `progess` option ensures we're piping the progress bar to the right
+    place, and not just always polluting sys.stderr.
+    """
+    regex = re.compile(
+        r"Insert into [\w\.]*\W*\(([\n ,\w]*)\)\W*Values\W*\((['\w, \&\n.:/\(\)!?$%*-+#]*)\);"
```
ArmaanT, line 138: Can you add a quick line explaining this regex?

backend/review/migrations/0005_auto_20200503_1820.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.2.12 on 2020-05-03 22:20
```
ArmaanT, line 1: Can you squash this migration as well?

backend/review/import_utils/import_to_db.py
davish: I tried originally, but the way we want to override properties isn't the same for if there's a matching user or a matching username. If a user matches, we don't want to update the name since the name probably came from Shibboleth. If the name matches then we do want to update the user reference.

PR #206, opened 2020-05-09
Refactor tests into top-level test directory

davish:
I need a drink.

This PR:
- [ch1699] Replaces usages of string interpolation + hardcoded URLs with [`reverse()`](https://docs.djangoproject.com/en/3.0/ref/urlresolvers/#reverse).
- [ch1999] Factors all tests into a top-level `tests` directory to make it easier to have multiple test files per re-usable app. When it made sense, I separated test files within apps into more discrete groups as well.


backend/.dockerignore
```diff
@@ -26,3 +26,4 @@ tests/
 # PCX
 pcr-backup
 frontend/
+tests/
```
ArmaanT, line 29: Tests are already in the dockerignore. Also technically everything in the Dev section and below can be removed without any issues since the images are only built in CircleCI

backend/tests/courses/test_api.py
```diff
@@ -0,0 +1,208 @@
+from django.test import TestCase
+from django.urls import reverse
+from options.models import Option
+from rest_framework.test import APIClient
+
+from courses.models import Department, Instructor, Requirement
+from courses.util import create_mock_data, get_or_create_course
```
ArmaanT, line 7: If `create_mock_data` is only used to generate data for tests, can we move it into a util file in the tests directory?

backend/alert/urls.py
```diff
@@ -13,8 +13,7 @@
 
 urlpatterns = [
     path("", views.index, name="index"),
-    path("courses/", courses.views.SectionList.as_view()),
-    path("statusupdate/<slug:full_code>/", StatusUpdateView.as_view()),
```
ArmaanT, line 17: Was statusupdate part of the 3rd party api that's getting removed soon?

backend/Pipfile
```diff
@@ -44,6 +44,8 @@ uritemplate = "*"
 uwsgi = "*"
 ddt = "*"
 tqdm = "*"
+nose = "*"
```
ArmaanT, line 47: What's nose being used for? It looks like you're just using Django's built-in test runner here

backend/alert/urls.py
davish: statusupdate was included in the url routes for both `alert` and `courses`. Just removed it from `alert` so it only lives in one place.

backend/Pipfile
davish: I was getting an error when running tests locally for some reason, it said nose wasn't installed. Adding this fixed that. :/

davish: ddt needs it, but doesn't include it in its dependencies, apparently.

PR #209, opened 2020-05-11
Remove django-celery-beat

davish:
[ch1985]

We were using django-celery-beat as a way to control cron jobs from django and spin off one-off tasks. We can replace one-off tasks with manage.py commands (work was duplicated there anyway) and replace the cron jobs with kubernetes cron jobs through Icarus. This will reduce our dependencies, but also more importantly there were a ton of weird dependency restrictions on both `django-celery-beat` and `django-celery-results`. By removing those, we should be able to update to Django 3.0.

These miscellaneous celery tasks were one of the big sore points for missing code coverage, so removing them also has a considerably positive impact on that metric.


k8s/values.yaml
```diff
@@ -42,3 +35,10 @@ applications:
       hosts:
         - host: penncoursealert.com
           paths: ["/"]
+
+cronjobs:
+  - name: load-courses
+    schedule: "0 3 * * *"
```
ArmaanT, line 41: Just confirming this is supposed to run at 3 am every day

backend/courses/management/commands/loadcourses.py
```diff
@@ -1,8 +1,13 @@
 import logging
```
ArmaanT, line 1: Do you think it also makes sense to have this check a runtime option so that it only runs during add/drop? That way we don't have to make any code changes to disable/enable it

k8s/values.yaml
davish: Yep

backend/courses/management/commands/loadcourses.py
davish: I was thinking about this but I think we want it to always run. We never know when people are going to be looking at courses, and it doesn't hurt to have the most up-to-date info.

PR #210, opened 2020-05-12
Consolidate Styled Components

davish:
[ch1963]

Move common styled components into `src/common/` directory. This is mostly a first-pass, but it looks like any other styled components are very component-specific.


eyingxuan:
LGTM! Merge when ready

PR #211, opened 2020-05-12
PCR Routes

davish:
[ch2024]

This PR will add API routes for accessing aggregations of PCR data needed for the PCR frontend.

Routes were modeled after the `/display` routes that are used in the current interation of the PCR frontend. Outputted JSON should be almost identical.


backend/review/annotations.py
```diff
@@ -0,0 +1,116 @@
...
+    """
+    if fields is None:
+        fields = ["course_quality", "difficulty", "instructor_quality", "work_required"]
+    queryset = queryset.annotate(
+        **{
+            (prefix + field): Subquery(
+                ReviewBit.objects.filter(field=field, **subfilters)
+                .values("field")
+                .order_by()
+                .annotate(avg=Avg("average"))
```
mureytasroc, line 50: I know this is old code but I'm wondering where the ReviewBit score field is referenced?  Also, maybe I missed this in the Django docs, but where is the field `average` coming from for `Avg("average")` to work?  Conceptually I get why you would want `Avg("average")` in this line to return the average score for the field group but I'm not sure why it would with this code...

backend/review/models.py
```diff
@@ -61,6 +61,37 @@ def get_averages(course_code, instructor_name=None, fields=None):
...
+    ("RRECOMMENDMAJOR", "Recommend Major", "recommend_major"),
+    ("RRECOMMENDNONMAJOR", "Recommend Non-Major", "recommend_nonmajor"),
+    ("RABILITIESCHALLENGED", "Abilities Challenged", "abilities_challenged"),
+    ("RCLASSPACE", "Class Pace", "class_pace"),
+    ("RINSTRUCTOREFFECTIVE", "Instructor Effectiveness", "instructor_effective"),
+    ("RNATIVEABILITY", "Native Ability", "native_ability"),
+)
+
+# Maps column name from SUMMARY sql tables to common slug.
+COLUMN_TO_SLUG = dict([(x[0], x[2]) for x in REVIEW_BIT_LABEL])
```
mureytasroc, line 89: Any reason why this isn't done with a dict comprehension?

backend/review/models.py
```diff
@@ -61,6 +61,37 @@ def get_averages(course_code, instructor_name=None, fields=None):
...
+# Maps "context" value from RATING table to common slug.
+CONTEXT_TO_SLUG = dict([(x[1], x[2]) for x in REVIEW_BIT_LABEL])
```
mureytasroc, line 91: See previous comment

backend/review/views.py
```diff
@@ -1,4 +1,201 @@
...
+                    "num_semesters": r["average_semester_count"],
+                }
+                for r in nested
+            },
+        }
+    )
+
+
+@api_view(["GET"])
+def course_reviews(request, course_code):
```
mureytasroc, line 60: This is more of a question than a comment -- would there be any benefit to eventually using a viewset for these routes?  How come we don't already?

backend/review/models.py
davish: no reason... will change!

backend/review/models.py
davish: ditto

backend/review/views.py
davish: ViewSets define REST actions that make a lot of sense when you're talking about a specific "resource" -- generally in our cases, these resources are individual django models that have serializers defined for them. With these aggregations, though, I couldn't find a way to fit the desired output format into a serializer that made much sense. There probably is a way to do it, but I felt like it'd be shoe-horned in so much that it made more sense to use "bare" ApiViews.

backend/review/annotations.py
davish: In the last PR, the `score` field got renamed to `average`. So `Avg("average")` is annotating the average of all the scores from all the filtered ReviewBits. The reason the field was renamed was because in the big PCR dump we have a lot more information on Reviews, including median, standard deviation, and distribution that we can all load in.

mureytasroc: That's awesome.  Great to hear we have that other data.

mureytasroc: Wait what would distribution be?

davish: We can see how many people voted 0, 1, 2, 3 and 4 for each field

PR #212, opened 2020-05-14
Frontend Infrastructure Overhaul

eyingxuan:
[ch1932]

This PR aims to introduce Prettier configurations/pre-commit hooks to the frontend codebase. To avoid style conflicts, we use ```eslint-config-prettier```, a package that contains community-gathered ESLint rules that may conflict with the vanilla prettier configuration. The implications of the Prettier integration are that there will be slight changes in style, as we will be defaulting to Prettier configurations now, outside of indentation and dangling commas, which are configured in ```prettierrc.json```.

As we move forward, we might find some Prettier defaults to be less appealing - we can manually fine tune the ```prettierrc.json``` configuration file then.

On the way, we discovered the magic that is Yarn Workspaces. This PR will also migrate CPX scripts to Yarn Workspaces so ```shared-components``` becomes a dependency for ```plan``` and ```alert```.


eyingxuan:
Yeap! I'm planning to add pre-commit hooks for prettier, and have ESLint do the final pass in CI. Hopefully that works out!

eyingxuan:
Actually, I just realized ESLint seems to not be catching some rules after extending prettier (such as semicolon after statement). I'll look into this more.

eyingxuan:
Never mind, eslint-plugin-prettier fixes this

PR #214, opened 2020-05-23
Next.js Migration for Penn Course Plan

eyingxuan:
[ch2032] 

sErVerSidErEnDerinG

TODOS:
- [x] Figure out deployment (probably set up env variable so next listens on port 80 in k8s)
- [x] ~~Login: looks like SSR really messes with current logic, so we might have to actually refactor accounts before it is fully functional~~ Forgot to proxy ```/accounts```
- [x] Visual glitches 
- [x] Investigate spurious localhost:3000 GET requests (suspect proxy and Hot Reloading is interfering)
- [x] Investigate miscompiled headers

Screen Shot 2020-05-25 at 6 07 44 PM


sigmachirality:
oui I thought you guys weren't doing this you hacks

davish:
Also nothing major's gonna have to change for deployment! You can see what clubs does here: https://github.com/pennlabs/penn-clubs/blob/master/k8s/values.yaml and here: https://github.com/pennlabs/penn-clubs/blob/master/frontend/Dockerfile. We can probably just copy the `Dockerfile` honestly.

eyingxuan:
Ayyyy LETS GO! Fingers crossed.

PR #216, opened 2020-05-30
Migrate PCA to Next

davish:
[ch2167]

Following @eyingxuan's example from PCP.


frontend/alert/Dockerfile
```diff
@@ -9,15 +9,11 @@ COPY package.json /app
 COPY yarn.lock /app
 COPY alert/ /app/alert
 COPY shared-components/ /app/shared-components
-RUN yarn workspace alert install
-RUN yarn workspace pcx-shared-components install
+RUN yarn workspace penncoursealert install --production --frozen-lockfile
```
eyingxuan, line 12: This name has to match the name field in frontend/alert/package.json, so if we want to keep this for consistency's sake, we will have to update the name in package.json

frontend/alert/package.json
```diff
@@ -24,13 +27,13 @@
     "styled-components": "^5.1.0"
```
eyingxuan, line 27: We should get rid of the react-scripts dependency above

frontend/alert/pages/index.js
```diff
@@ -4,17 +4,15 @@ import PropTypes from "prop-types";
 import ReactGA from "react-ga";
 import * as Sentry from "@sentry/browser";
 
-import "./App.css";
```
eyingxuan, line 7: We're missing some CSS from App.css that was deleted and not replaced

eyingxuan:
I also forgot to pass the "pathname" props, which should be window.location.pathname in the current pages/index.js to AccountIndicator - this currently does not affect the App, since the Login button on the navbar is blocked by the modal, but just in case the UI changes for login, we should probably fix that.

eyingxuan:
FA icon in the footer doesn't work currently - need to include FA css

eyingxuan:
Also a reminder to change .dockerignore to discard .next and change .lintstaged.yml to pick up ts and tsx files

eyingxuan:
Had some issues with styled-components and SSR for plan, found this that seems to work the first 5 minutes I've tried - should probably add to alert too. 

https://dev.to/aprietof/nextjs--styled-components-the-really-simple-guide----101c

eyingxuan:
Looks like I'm still getting a className did not match warning, but because the styles are injected, things still look fine. I'll investigate more

PR #224, opened 2020-09-07
Add check for SEMESTER option, fix PCA old registrations bug

mureytasroc:


mureytasroc:
Not sure how I could write unit tests for the raising of an exception if get_value is None... I tested this manually in my dev environment and it works.  Is it ok to merge even with the lower code cov as a result of this?

The result of my tests (tested for both plan/schedules and alert/registrations):

```
HTTP 500 Internal Server Error
Allow: GET, POST
Content-Type: application/json
Vary: Accept
{
    "detail": "The SEMESTER runtime option is not set.  If you are in dev, you can set this option by running the command 'python manage.py setoption -key SEMESTER -val 2020C', replacing 2020C with the current semester, in the backend directory (remember to run 'pipenv shell' before running this command, though)."
}
```

davish (approved):
Looks good to me! I think it should be possible to write a test for this, though... The JSON that you showed in your local tests can be found in `response.data` just as normal, and `response.status_code` should be 500.
I know that the semester is set in the `setUp` method, so you could either un-set the option in the individual test (delete the instance) or create a new `TestCase` subclass without the `setUp()` method.

mureytasroc:
I made a bunch more changes (basically replacing `get_value("SEMESTER")` with `get_current_semester` everywhere except for in courses/views.py (which seems to handle a missing SEMESTER option in a controlled way):

```
def get_semester(self):
        semester = self.kwargs.get("semester", "current")
        if semester == "current":
            semester = get_value("SEMESTER", "all")

        return semester
```

So I think you should probably review again, just to make sure I'm not messing up anything with these changes.  All tests are passing though, and I added tests to ensure a 500 is returned when the semester is not set.

PR #229, opened 2020-09-28
Add types to package.json

haoyxc:
Add @types/react to the deps

### DIFF
--- a/frontend/plan/package.json
+++ b/frontend/plan/package.json
@@ -15,7 +15,11 @@
   "dependencies": {
     "@material-ui/core": "^4.5.2",
     "@sentry/browser": "^5.19.2",
+    "@types/react": "^16.9.49",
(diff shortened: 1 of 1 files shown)


frontend/plan/package.json
```diff
@@ -15,7 +15,11 @@
   "dependencies": {
     "@material-ui/core": "^4.5.2",
     "@sentry/browser": "^5.19.2",
+    "@types/react": "^16.9.49",
+    "@types/react-dom": "^16.9.8",
     "@types/styled-components": "^5.1.3",
+    "add": "^2.0.6",
```
davish, line 21: were these (`add`, `and`) on purpose? Also don't know if `yarn` needs to be added either

davish (changes requested):

PR #231, opened 2020-09-28
Refactor PCR Views

davish:
[ch2417]
Refactor PCR views for clarity, move helper functions to a separate file, and generally work towards making this file maintainable.

Still need to decide if we want to move all instances of `annotate_average_and_recent` to `annotate_reviews`. The former does aggregation in-database, and the latter aggregates in Python from flat review objects. The python aggregation is used when the SQL aggregations were too slow, but we might want to just simplify it and make everything go through Python for consistency's sake.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,3 +13,31 @@ installed.
 ## API Documentation
 API Docs can be found at `/api/documentation` on the back-end server. Also check out the code for more explanations
 and documentation! We've tried to keep it up-to-date.
+
+## Runbook
+This section is to collect thoughts/learnings from the codebase that have been hard-won, so we don't lose a record of it
+if and when the information proves useful again
+
+### Derived fields
+Normally, derived fields on models are represented as `@property`-decorated functions. However, there are a few in
+the codebase that need to be accessed on the database layer without `JOIN`s. So that they can be indexed.
+Specifically, these are the `full_code` fields on `Course` and `Section` models, which are derived from fields on related
+models.
+
+These are updated every time the `save()` method is called. However, it's possible to get into a state 
+(such as with db migrations) where `full_code` isn't set properly.
+
+Open a shell in production, and run this small script:
+```python
+from tqdm import tqdm
+from courses.models import Section, Course
+
+for c in tqdm(Course.objects.all().select_related("department")):
+    c.save()
+
+for s in tqdm(Section.objects.all().select_related("course")):
+    s.save()
+```
+
+`tqdm` will give you a nice progress bar as the script completes. The `select_related` clause speeds up the query,
+avoiding what would be a pretty nasty N+1 scenario.
--- a/backend/review/util.py
+++ b/backend/review/util.py
@@ -1,4 +1,5 @@
 import re
+from typing import Dict, List
 
 
 def titleize(name):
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -13,3 +13,31 @@ installed.
...
+```python
+from tqdm import tqdm
+from courses.models import Section, Course
+
+for c in tqdm(Course.objects.all().select_related("department")):
+    c.save()
+
+for s in tqdm(Section.objects.all().select_related("course")):
+    s.save()
+```
```
sigmachirality, line 40: do you need to activate a virtual env to run this in prod or nah, not familiar with how y'all got everything set up tbh

backend/review/util.py
```diff
@@ -18,3 +19,87 @@ def titleize(name):
...
+    Turns fields from python snake_case to the PCR frontend's rCamelCase.
+    """
+    return "r" + "".join([x.title() for x in s.split("_")])
+
+
+def make_subdict(field_prefix, d):
+    """
+    Rows in a queryset that don't represent related database models are flat. But we want
+    our JSON to have a nested structure that makes more sense to the client. This function
+    takes fields from a flat dictionary with a certain prefix
```
sigmachirality, line 35: takes fields from a flat dictionary...and what?

backend/review/util.py
```diff
@@ -18,3 +19,87 @@ def titleize(name):
...
+def dict_average(entries: List[Dict[str, float]]) -> Dict[str, float]:
+    """
+    Average a list of dicts into one dict with averages.
+    :param entries:
+    :return:
+    """
+    keys = []
+    for entry in entries:
+        keys.extend(entry.keys())
+    keys = set(keys)
```
sigmachirality, line 54: nit, but can you start with a set and use .update instead of .extend on a list?

backend/review/util.py
sigmachirality: and filters/formats their keys is what I think you were going to write

backend/review/util.py
```diff
@@ -18,3 +19,87 @@ def titleize(name):
...
+        recent_scores = [r["scores"] for r in reviews if r["semester"] == latest_sem]
+        aggregated[k] = {
+            "id": k,
+            "average_reviews": dict_average(all_scores),
+            "recent_reviews": dict_average(recent_scores),
+            "latest_semester": latest_sem,
+            "num_semesters": len(set([r["semester"] for r in reviews])),
+            **{
+                response_prop: reviews[0][response_prop]
+                for response_prop, _ in extra_fields.items()
```
sigmachirality, line 101: nit: would it make more sense to name response_prop extra_field here?

backend/review/util.py
```diff
@@ -18,3 +19,87 @@ def titleize(name):
...
+    """
+    Aggregate a list of reviews, grouping by some field.
+    :param reviews: A list of dictionaries representing Review objects, with reviewbits inlined
+    using review.annotations.review_averages(). And dict-ified by calling .values() on a queryset.
+    :param group_by: Field to group by in the review.
+    :param extra_fields: Any extra fields from the dictionaries to carry through to the response.
+    :return: Average reviews, recent reviews, number of semesters taught, and other data needed
+    for the response to the frontend.
+    """
+    grouped_reviews = dict()
```
sigmachirality, line 75: might be good to add comments for each pass, like "group reviews by field using dict" and then "aggregate reviews by group_by"

sigmachirality (approved):
Some nits but reading through the rewritten code makes me appreciate abstractions. 

It's great to see how the readability of api views has been improved, especially comparing to some of the views I wrote in pcr freshman year lol

README.md
davish: Nope! No virtualenv in prod. @pawalt probably should write a user guide for Bastion (the system for sshing into prod), but I know that it's going thru a lot of changes this semester

pawalt: This is dockerfile-dependent. If you have the `--system` flag on your pipenv install, then you don't need a venv, otherwise you have to do a `pipenv run ./manage.py`. Bastion is changing but this will not change.

PR #232, opened 2020-10-01
Mobile alert notifications

mureytasroc:


backend/Pipfile
```diff
@@ -18,7 +18,6 @@ django-debug-toolbar = "*"
 
 [packages]
 django = "*"
-psycopg2 = "*"
```
mureytasroc, line 21: Make sure to undo this change before merging

backend/alert/views.py
```diff
@@ -83,22 +83,22 @@ def accept_webhook(request):
...
+    alert_for_course_called = False
+
     should_send_alert = (
         get_bool("SEND_FROM_WEBHOOK", False)
-        and course_status == "O"
+        and (course_status == "O" or course_status == "C")
         and get_value("SEMESTER") == course_term
     )
 
-    alert_for_course_called = False
```
mureytasroc, line 92: Why delete this?

backend/alert/views.py
```diff
@@ -83,22 +83,22 @@ def accept_webhook(request):
...
         try:
-            alert_for_course(course_id, semester=course_term, sent_by="WEB")
+            alert_for_course(course_id, semester=course_term, sent_by="WEB", course_status=course_status)
             alert_for_course_called = True
             response = JsonResponse({"message": "webhook recieved, alerts sent"})
         except ValueError:
             response = JsonResponse({"message": "course code could not be parsed"})
-    else:
-        response = JsonResponse({"message": "webhook recieved"})
+    response = JsonResponse({"message": "webhook recieved"})
```
mureytasroc, line 101: Why remove the else?

mureytasroc (commented):
Generally looks good (I made a bunch of changes to make the whole close_notification thing more robust / fix all failing test cases; it turns out that is not a trivial feature to implement).  We definitely need some test cases though before we can merge (I can write some for the close_notification thing but you should def write some for push_notifications if possible).

backend/alert/alerts.py
```diff
@@ -48,46 +49,134 @@ def __init__(self, template, reg):
...
-    def send_alert(self):
+    def send_alert(self, close_notification=False):
         pass
 
 
 class Email(Alert):
     def __init__(self, reg):
-        super().__init__("alert/email_alert.html", reg)
+        super().__init__(
+            "alert/email_alert.html", reg, close_template="alert/email_close_alert.html"
```
davish, line 72: Did we decide on sending email close alerts? Not a cost issue, but thought we were keeping close alerts to mobile for now as an exclusive feature.

backend/alert/alerts.py
```diff
@@ -48,46 +49,134 @@ def __init__(self, template, reg):
...
             email = self.registration.email
         else:
             return False
+
         try:
+            if close_notification:
+                if not self.close_text:
+                    # This should be unreachable
+                    return None
+                alert_subject = "RE: %s is now open!" % self.registration.section.full_code
```
davish, line 92: Small thing, but if we want to make sure we're replying to the right thread, should probably pull out `%s is now open!` into its own string so we know it's the same.

backend/alert/alerts.py
```diff
@@ -48,46 +49,134 @@ def __init__(self, template, reg):
...
+                alert_text = self.close_text
+            else:
+                alert_subject = "%s is now open!" % self.registration.section.full_code
```
davish, line 95: Should probably be using f-strings here and above too

backend/alert/alerts.py
```diff
@@ -48,46 +49,134 @@ def __init__(self, template, reg):
...
+        and None if notification was attempted to be sent but an error occurred.
+        """
+        if (
+            self.registration.user
+            and self.registration.user.profile
+            and self.registration.user.profile.push_notifications
+        ):
+            # Only send push notification if push_notifications is enabled
+            pennkey = self.registration.user.username
+            bearer_token = str(10101)
```
davish, line 157: Should pull this out into an environment variable (see the bottom of `settings/base.py` for lots of examples

backend/alert/alerts.py
```diff
@@ -48,46 +49,134 @@ def __init__(self, template, reg):
...
+                    # This should be unreachable
+                    return None
+                alert_title = "%s just closed." % self.registration.section.full_code
+                alert_body = self.close_text
+            else:
+                alert_title = "%s is now open!" % self.registration.section.full_code
+                alert_body = self.text
+            try:
+                response = requests.post(
+                    "https:/api.pennlabs.org/notifications/send/internal",
```
davish, line 169: Not sure if this needs to be in this PR, but we should probably hunt around for all of the hardcoded URLs and put them in a `constants.py` file or something

backend/alert/models.py
```diff
@@ -279,28 +314,99 @@ def save(self, *args, **kwargs):
...
+            "close_notification_sent": False,
+        }
+
+    @property
+    def is_waiting_for_close(self):
+        """
+        Returns True iff the registration would send a close notification
+        when the watched section changes to closed
+        """
+        for k, v in self.is_waiting_for_close_filter().items():
```
davish, line 356: if this function is the only place the above dict is used, should probably move this inline.

backend/alert/tasks.py
```diff
@@ -25,20 +25,34 @@ def run_course_updates(semester=None):
...
+    elif course_status == "C":
+        # Use the is_waiting_for_close_filter filters statically defined in the Registration model
+        return list(section.registration_set.filter(**Registration.is_waiting_for_close_filter()))
+    else:
+        return []
 
 
 @shared_task(name="pca.tasks.send_course_alerts")
-def send_course_alerts(course_code, semester=None, sent_by=""):
+def send_course_alerts(course_code, semester=None, sent_by="", course_status="O"):
```
davish, line 53: Putting a default status here makes me kind of nervous, since this is data we should be relying on the webhook for. It might be a bit more work to set a variable at every call site, but I think it would be worth it to make sure bugs don't pop up.

backend/alert/tasks.py
```diff
@@ -25,20 +25,34 @@ def run_course_updates(semester=None):
...
     if semester is None:
         semester = get_current_semester()
 
-    for reg in get_active_registrations(course_code, semester):
-        send_alert.delay(reg.id, sent_by)
+    for reg in get_registrations_for_alerts(course_code, semester, course_status=course_status):
```
davish, line 57: I think this is a really elegant solution to the problem! Is there any code that still uses `get_active_registrations` or can that code safely be deleted?

backend/alert/views.py
```diff
@@ -28,8 +28,8 @@
 logger = logging.getLogger(__name__)
 
 
-def alert_for_course(c_id, semester, sent_by):
-    send_course_alerts.delay(c_id, semester=semester, sent_by=sent_by)
+def alert_for_course(c_id, semester, sent_by, course_status="O"):
```
davish, line 31: Seconded here, about defaulting to "O" for the status.

backend/courses/models.py
```diff
@@ -447,8 +447,8 @@ class StatusUpdate(models.Model):
         help_text="Was an alert was sent to a User as a result of this status update?"
     )
     # ^^^ alert_sent is true iff alert_for_course was called in accept_webhook in alert/views.py
-    # equivalently, iff SEND_FROM_WEBHOOK == True and SEMESTER == course_term, or the request
-    # is otherwise invalid
+    # equivalently, iff SEND_FROM_WEBHOOK == True and SEMESTER == course_term, and the request
```
davish, line 450: we love demorgan's law

backend/alert/alerts.py
```diff
@@ -48,46 +49,134 @@ def __init__(self, template, reg):
...
+        """
+        Returns False if notification was not sent intentionally,
+        and None if notification was attempted to be sent but an error occurred.
+        """
+        if (
+            self.registration.user
+            and self.registration.user.profile
+            and self.registration.user.profile.push_notifications
+        ):
+            # Do not send text if push_notifications is enabled
```
davish, line 122: I think we'll need some more checking eventually to handle the case where users log out of penn mobile on their phones, how do we re-enable texts? but this can be a conversation with mobile at some point.

backend/tests/alert/test_alert.py
```diff
@@ -69,37 +71,281 @@ def override_delay(modules_names, before_func, before_kwargs):
...
         self.assertTrue(mock_email.called)
         self.assertTrue(mock_text.called)
-        self.assertTrue(Registration.objects.get(id=self.r.id).notification_sent)
-        self.assertEqual("ADM", Registration.objects.get(id=self.r.id).notification_sent_by)
+        self.assertFalse(mock_push_notification.called)
+        r_legacy = Registration.objects.get(id=self.r_legacy.id)
+        self.assertTrue(r_legacy.notification_sent)
+        self.assertEqual("ADM", r_legacy.notification_sent_by)
+
+    def send_alert_helper(self, mock_email, mock_text, mock_push_notification, push_notification):
```
davish, line 106: some documentation on what this helper is doing would be useful!

backend/tests/alert/test_alert.py
```diff
@@ -69,37 +71,281 @@ def override_delay(modules_names, before_func, before_kwargs):
...
+        self.send_alert_helper(
+            mock_email, mock_text, mock_push_notification, push_notification=True
+        )
+
+    def test_send_alert(self, mock_email, mock_text, mock_push_notification):
+        self.send_alert_helper(
+            mock_email, mock_text, mock_push_notification, push_notification=False
+        )
+
+    def send_close_notification_helper(
```
davish, line 158: Again, docs on what the helper is doing would be great. Seems to be doing a lot that's hard to parse at first glance!

backend/tests/alert/test_alert.py
```diff
@@ -69,37 +71,281 @@ def override_delay(modules_names, before_func, before_kwargs):
...
+        self.send_close_notification_helper(
+            mock_email,
+            mock_text,
+            mock_push_notification,
+            push_notification=False,
+            auto_resubscribe=False,
+            manual_resubscribe=True,
+        )
+
+    def dont_resend_alert_helper(
```
davish, line 269: Ditto again. Will stop commenting but yeah, docstrings on helper functions would be amazing!

backend/tests/alert/test_alert.py
```diff
@@ -773,144 +1064,264 @@ def test_ignore_fields_phone_update(self):
...
         response = self.client.get(reverse("user-profile"))
         self.assertEqual(200, response.status_code)
         self.assertEqual(response.data["profile"]["phone"], "+12121234567")
         self.assertEqual(response.data["profile"]["email"], "example@email.com")
+        self.assertTrue(response.data["profile"]["push_notifications"])
+        self.assertEqual(response.data["username"], "jacob")
+        self.assertEqual(response.data["first_name"], "fname")
+        self.assertEqual(response.data["last_name"], "lname")
+
+    def test_ignore_fields_push_notifications_update(self):
```
davish, line 1080: Why are we ignoring updates to the push settings? seems like users should be able to change that from the client?

davish (changes requested):
Really awesome work! Some small comments, and requests for a bit more documentation in parts. But should be good to go after those!

backend/alert/alerts.py
mureytasroc: So I was assuming standardization on the backend would be better and then we could regulate whether or not to afford the feature on the frontend (since close alerts are disabled by default, if the frontend doesn't change they won't be available to users on the web version).  Does this make sense or should I remove on the backend?

backend/alert/models.py
mureytasroc: It is also used in alert/tasks.py (line 47), to filter registrations (since you can't filter on a property):

`return list(section.registration_set.filter(**Registration.is_waiting_for_close_filter()))`

backend/alert/tasks.py
mureytasroc: Good point, it is no longer being used by anything.  Just deleted `get_active_registrations`

backend/tests/alert/test_alert.py
mureytasroc: Sorry, I should really document these tests better.  This is testing the functionality that if the user only updates push_notifications, leaving out other fields, it ignores the other fields and only updates push_notifications.  I should really change the PCA PUT route to PATCH (have been meaning to do this)

backend/alert/alerts.py
mureytasroc: Agreed

backend/alert/tasks.py
mureytasroc: Fixed

backend/alert/alerts.py
mureytasroc: That's a good point; maybe we make enabling one automatically disable the other?

backend/alert/alerts.py
mureytasroc: Reached out to mobile backend to ask what the default bearer token should be... will move to an environment variable

backend/alert/alerts.py
```diff
@@ -48,46 +49,134 @@ def __init__(self, template, reg):
...
+                if response.status_code != 200:
+                    logger.exception(
+                        f"Push Notification {response.status_code} Response: {response.content}"
+                    )
+                    return None
+            except requests.exceptions.RequestException as e:
+                logger.exception(f"Push Notification Request Error: {e}")
+                return None
+            return True
+        return False
```
r3khaan, line 182: Instead of returning `False` in line 182, shouldn't we be returning `None` since `self.registration.user` is `None` due to some error? (i.e. we do want to send a push notif, but there is an error preventing us from doing so)

backend/alert/templates/alert/push_notif.txt
```diff
@@ -0,0 +1,2 @@
+{{course}} is now open! Register on PennInTouch: http://bit.ly/pitouch. {% if auto_resubscribe %}To cancel alerts, visit{% else %}Resubscribe at{% endif %} http://penncoursealert.com#manage
+ - Thanks for using Penn Course Alert!
```
r3khaan, line 2: I'm pretty sure that users can't click on individual links in mobile push notifications. Ideally, when we roll out the feature, the notification will have actions (accessed via the "...") like "go to pennintouch", "cancel alert", "resubscribe for alert", "manage alert in pennmobile".

We don't need to change it now, because I think it will remind us to add actions later, but just thought I'd give the heads up.

r3khaan (approved):
LGTM! Good stuff. Excited to roll this out soon.

backend/alert/alerts.py
mureytasroc: So this is actually behavior which is grandfathered in from the old PCA; self.registration.user is None on legacy registrations, so in that case it just wouldn't send any push notifications.  Not really an error state so I don't think returning None makes sense, but it's kind of an unreachable case now since there is no way of creating legacy registrations anymore.

backend/alert/alerts.py
mureytasroc: Maybe if push notifications are enabled we grey out the phone number box on the web frontend with some sort of button to disable push / enable text

astrike30: I think it might be worth emphasizing on the sign-up screen if alerts are going to the app or via text. The push notification token should still be good even if they get logged out of the app, and if they delete the app I think the next time they sign-up for a class they would know to change it in most cases?

backend/alert/templates/alert/push_notif.txt
astrike30: Not sure how this works on the mobile side, but does the notification go straight to pennintouch when tapped?

astrike30 (approved):
All looks good to me 👍  hopefully nothing breaks in production but I think features like this (sending notifications etc) are prone to being unstable at first. The docs are incredible btw!

backend/PennCourses/docs_settings.py
```diff
@@ -320,9 +335,39 @@
         is a registration created by resubscribing to the previous registration (once that
         registration had triggered an alert to be sent), either manually by the user or
         automatically if auto_resubscribe was set to true.  Then, it follows that the head of the
-        resubscribe chain is the most relevant registration for that section; if any of the
-        registrations in the chain are active, it would be the head.  And if the head is active,
-        none of the other registrations in the chain are active.
+        resubscribe chain is the most relevant Registration for that user/section combo; if any
```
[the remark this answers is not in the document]
astrike30: oh I did add a few comments above ^^ but I wasn't in review mode 🤦🏻‍♂️ just wanted to make sure everyone sees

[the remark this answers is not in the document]
mureytasroc: So docs_settings.py actually is for the frontend docs.  Added your suggestions in the backend docs; check them out at `http://localhost:8000/admin/doc/models/alert.registration/` when the backend server is running.

backend/alert/alerts.py
davish: Should be fine to keep it in as long as we make sure to test that case

backend/alert/alerts.py
davish: Let's follow up on this with @ArmaanT and @r3khaan. Not sure how much of an issue it will be.

backend/alert/alerts.py
```diff
@@ -48,46 +51,127 @@ def __init__(self, template, reg):
...
+            return False
         if self.registration.user is not None and self.registration.user.profile.phone is not None:
             phone_number = self.registration.user.profile.phone
         elif self.registration.phone is not None:
             phone_number = self.registration.phone
         else:
             return False
 
-        return send_text(phone_number, self.text)
+        if close_notification:
```
davish, line 130: We should never be sending close notifications via text messages -- this should be `return None` whether or not `close_text` is available.

davish (approved):
One pretty major comment (ALWAYS return `False` when trying to send a close notification for a text message, never even try to send) for spam and cost reasons. Otherwise, the other outstanding comment is how to handle the silent logout scenario on mobile. We need to know if we can still send push notifications if the user has happened to be logged out in penn mobile (because device ID is still stored? not sure how the API server works here) want to pull in @r3khaan and @ArmaanT to double-check

r3khaan:
> One pretty major comment (ALWAYS return `False` when trying to send a close notification for a text message, never even try to send) for spam and cost reasons. Otherwise, the other outstanding comment is how to handle the silent logout scenario on mobile. We need to know if we can still send push notifications if the user has happened to be logged out in penn mobile (because device ID is still stored? not sure how the API server works here) want to pull in @r3khaan and @ArmaanT to double-check

We can send push notifs to users when they are logged out! The only case sending notifications fails is when the user deletes the app. Not sure what the protocol is for this case, but one idea is to send a request from mobile to backend confirming that the notif was received after every notification .. and if the request is not received, then we know to switch notif style from push notifs back to text notifs.

PR #233, opened 2020-10-03
Automatically Generate 403 Responses

mureytasroc:


backend/PennCourses/docs_settings.py
```diff
@@ -866,53 +866,86 @@ def _get_responses(self, path, method):
...
         ):
             if path in self.override_schema and method in self.override_schema[path].keys():
                 return {
+                    **{
+                        "200": {
+                            "description": "",
+                            **(
+                                {
+                                    "content": {
+                                        ct: {"schema": self.override_schema[path][method][200]}
```
davish, line 889: I'm guilty of this too, but could you spell out what this variable is? one and two letter variable names can get confusing.

davish: looks like it's `content_type`

backend/PennCourses/docs_settings.py
```diff
@@ -866,53 +866,86 @@ def _get_responses(self, path, method):
...
+                        # TODO: put something meaningful into it
                         "description": "",
+                    }
+                },
+                **auth_response_dict,
+            }
+        if path in self.override_schema and method in self.override_schema[path].keys():
+            return {
+                **{
+                    str(k): {
```
davish, line 920: same here. could you replace `k` and `v` with something a bit more specific? `k` seems to be the status code and `v` seems to be the message.

backend/PennCourses/docs_settings.py
```diff
@@ -866,53 +866,86 @@ def _get_responses(self, path, method):
...
                                     for ct in self.response_media_types
                                 }
-                                if 200 in self.override_schema[path][method].keys()
+                                if k in self.override_schema[path][method].keys()
                                 else {
                                     ct: {"schema": response_schema}
                                     for ct in self.response_media_types
                                 }
                             }
+                            if "[SCHEMA]" in v
```
davish, line 934: This should be documented -- ideally there'd be a better way to do this than a string keyword, but this is fine as long as it's documented (both at the top [ where i think it is ] and right here.

davish (commented):
some small naming things, but after that looks good to go!!!

backend/PennCourses/docs_settings.py
mureytasroc: This naming convention was from the DRF AutoSchema code but you're definitely right that `content_type` is better.

backend/PennCourses/docs_settings.py
mureytasroc: Done.  Good point, will make sure to use more descriptive names in the future.

backend/PennCourses/docs_settings.py
mureytasroc: It is documented at the top, and I documented it here as well.

davish (approved):
Let's rock!!

PR #235, opened 2020-10-22
Migrate Selector to TS

davish:
Still need to test things out and make sure nothing got broken on the frontend. React-redux higher-order components really don't seem to like typescript at all, so there's some weird tsignores around the connect call. We should probably move over to redux hooks anyway.

- Selector migration
- add react redux types to package.json
- add `tsc --noemit` to the lint step


frontend/plan/components/selector/CourseInfo.tsx
```diff
@@ -1,13 +1,12 @@
...
-import PropTypes from "prop-types";
 import styled from "styled-components";
 import CourseDetails from "./CourseDetails";
 import SectionList from "./SectionList";
 import { Course } from "../../types";
 
 interface CourseInfoProps {
     course: Course;
-    back: () => void;
+    back?: () => void;
```
haoyxc, line 9: just for my own clarification -- does this mean back is option? what's the reason it's optional?

frontend/plan/components/selector/TagList.tsx
```diff
@@ -26,7 +29,7 @@ export default function TagList({
                     className="taglist"
                     style={{ height: expanded ? "auto" : 0 }}
                 >
-                    {hiddenTags.map((elt) => (
+                    {hiddenTags.map((elt, i) => (
```
haoyxc, line 32: what was the reason for adding i? do you need it in typescript?

haoyxc (commented):
sorry this took a while, but lgtm! this only thing is (and this is definitely not specific to this PR) but  this code could probably use a little more documentation?

haoyxc (approved):

frontend/plan/components/selector/CourseInfo.tsx
davish: Yep! This means that the `back` property is either `() => void` or `undefined`, which is basically missing. If there's only one course in the search results, we automatically open its course details and don't provide a back button. The CourseInfo component then isn't passed a function to go back, and so won't show a back button.

frontend/plan/components/selector/TagList.tsx
davish: i actually think i don't need this here :P thanks for pointing it out!

frontend/plan/components/selector/CourseDetails.tsx
```diff
@@ -1,12 +1,30 @@
...
-import PropTypes from "prop-types";
 import TagList from "./TagList";
 import Badge from "../Badge";
 import ShowMore from "../ShowMore";
+import { Course, School } from "../../types";
 
-const getReqCode = (school, name) =>
-    `${{ SAS: "C", SEAS: "E", WH: "W" }[school]}: ${name}`;
-const annotatePrerequisites = (text, onClick) => {
+const schoolToLetter = (school: School) => {
```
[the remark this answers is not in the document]
davish: It's definitely general, but we don't need to condense schools to a single character anywhere else, so I feel like it's fine keeping it in here until we do need to (if we do at some point)

PR #236, opened 2020-10-23
Migrate schedule to TS

jeffrey-xiao-17:
[ch2797]
Changes:
- Schedule migration
- Style components for Stats.tsx

Remaining TODOS/FIXME's:
1. Reconsider "any" types in Schedule.tsx (placed temporarily for testing)


frontend/plan/components/schedule/MyCircularProgressBar.tsx
```diff
@@ -24,7 +28,7 @@ export default function MyCircularProgressBar(props) {
     return (
         <CircularProgressbar
             value={(value / 4) * 100}
-            strokeWidth="14"
+            strokeWidth={14} // Changed from "14" to {14} to fix type string not assignable
```
davish, line 31: Don't think this comment is necessary

frontend/plan/components/schedule/Schedule.tsx
```diff
@@ -91,8 +118,8 @@ class Schedule extends Component {
                 "indigo",
             ];
             // some CIS120: `used` is a *closure* storing the colors currently in the schedule
-            let used = [];
-            return (c) => {
+            let used: string[] = [];
+            return (c: string) => {
```
davish, line 122: So if you look in `SchedBlock.tsx`, you'll see a `Color` enum or something to that effect. You should pull that enum out into the types file and then change all these strings and string arrays dealing with color into references to that enum. Should make it easier to add new colors later on, and ensures we don't have any typos!

frontend/plan/components/schedule/Schedule.tsx
```diff
@@ -142,7 +183,20 @@ class Schedule extends Component {
             // for every conflict of size k, make the meetings in that conflict
             // take up (100/k) % of the square, and use `left` to place them
             // next to each other.
-            const group = Array.from(conflict.values());
+            const group: {
+                day: Day;
```
davish, line 187: This type should probably have a name -- maybe `MeetingBlock` or something like that?

davish (changes requested):
Great work! Some smaller comments, main thing is the color enum and then this should be good to go!

frontend/plan/components/schedule/Schedule.tsx
jeffrey-xiao-17: Moved `Color` enum into types, and then edited Block and Schedule to use the new `Color` type rather than the string.

frontend/plan/components/schedule/Schedule.tsx
jeffrey-xiao-17: Extracted this type and placed it in `types.ts` under `MeetingBlock`

frontend/plan/components/Cart.tsx
```diff
@@ -0,0 +1,143 @@
...
+    courseInfo: (id: string) => void;
+    courseInfoLoading: boolean;
+    setTab: (_: number) => void;
+    lastAdded: {
+        id: string;
+    };
+    mobileView: boolean;
+}
+
+// Cart.propTypes = {
```
davish, line 22: remember to delete commented out code! cis120 style guide :P

frontend/plan/types.ts
```diff
@@ -57,6 +69,21 @@ export interface Meeting {
     room: string;
 }
 
+export interface MeetingBlock {
```
davish, line 72: a comment here about what this is would be helpful :)

davish (approved):
Looks good! one documentation nit but after that merge away :D

davish:
I think that you forgot to delete `Cart.js` so just make sure to do that before merge!

PR #237, opened 2020-10-24
Migrate most of search component to ts 

haoyxc:
everything except SearchBar.tsx

also changed the filters to booleans instead of 0/1s


frontend/plan/actions/index.js
```diff
@@ -211,7 +211,7 @@ function buildCourseSearchUrl(filterData) {
     const reqs = [];
     if (filterData.selectedReq) {
         for (const key of Object.keys(filterData.selectedReq)) {
-            if (filterData.selectedReq[key] === 1) {
+            if (filterData.selectedReq[key]) { // change from == 1
```
davish, line 214: No need to add comments with what changed :)
```suggestion
            if (filterData.selectedReq[key]) {
```

frontend/plan/actions/index.js
```diff
@@ -267,7 +267,7 @@ function buildCourseSearchUrl(filterData) {
             const applied = [];
             Object.keys(filterData[checkboxFields[i]]).forEach((item) => {
                 // eslint-disable-line
-                if (filterData[checkboxFields[i]][item] === 1) {
+                if (filterData[checkboxFields[i]][item]) { // changed from ===1
```
davish, line 270: ditto here

frontend/plan/components/search/CheckboxFilter.tsx
```diff
@@ -0,0 +1,94 @@
...
+                                </label>
+                            </div>
+                        </div>
+                    );
+                })}
+        </div>
+    );
+}
+
+CheckboxFilter.propTypes = {
```
davish, line 88: remember to remove proptypes after converting files

frontend/plan/components/search/SchoolReq.tsx
```diff
@@ -1,24 +1,43 @@
...
+
+interface SchoolReqProps {
+    startSearch: (searchObj: object) => void;
+    filterData: FilterData;
+    schoolReq: { [K in School]: Requirement[] };
+    addSchoolReq: (s: string) => void;
+    remSchoolReq: (s: string) => void;
+}
+
+type SchoolDisplay = "College" | "Engineering" | "Nursing" | "Wharton";
```
davish, line 13: Is there an enum that could be used for this type? would help with the repitition between here and the actual list.

frontend/plan/components/search/SchoolReq.tsx
```diff
@@ -1,24 +1,43 @@
...
+    ];
+    const [selSchool, setSelSchool] = useState<SchoolDisplay>("College");
 
-    const schoolCode = new Map();
-    schoolCode.set("College", "SAS");
-    schoolCode.set("Engineering", "SEAS");
-    schoolCode.set("Wharton", "WH");
-    schoolCode.set("Nursing", "NURS");
+    const schoolCode: { [K in SchoolDisplay]: School } = {
+        College: School.COLLEGE,
```
davish, line 31: @eyingxuan can correct me if I'm wrong, but I believe that if you give enums a string value, you can just cast the string to the enum type and things will work as expected.

frontend/plan/components/search/SearchBar.tsx
```diff
@@ -0,0 +1,450 @@
+// import React, { useEffect, useState } from "react";
```
davish, line 1: i would delete this file in this PR if it's 100% commented out and you've got another PR open

davish (approved):
Looks great! Some small changes, but after that go for the merge!!!

frontend/plan/components/search/SchoolReq.tsx
eyingxuan: I think the issue here was that the backend requires using an abbreviation of each school in the route (such as "SAS", "SEAS" etc) so School.COLLEGE is set to be equal to the abbreviated string, while we need what's displayed to be an actual string. 

What could be cleaner in retrospect is actually to write a function that turns the School enum into a display string, and use the enum itself in the useState to change between school requirements. What do you think? @davish @haoyxc

haoyxc: i'll defer to you guys since you probably know the codebase better than I do :-)

PR #238, opened 2020-10-25
Bulk Registration & Multi Registration

tanyew3st:
[ch2795] 

First pull request for the bulk registration.

TODO: update the endpoint to match design.


backend/alert/views.py
```diff
@@ -229,6 +235,249 @@ class RegistrationViewSet(AutoPrefetchViewSetMixin, viewsets.ModelViewSet):
...
+        {
+            "courses": ["CIS-120-001", "CIS-120-002"],
+            "auto_resubscribe": true
+        }
+
+        The response is going to be a dictionary that contains has the
+        name of the course as the key and the response message/data as the body.
+        """
+        # creating a list that will hold the response codes
+        response_codes = {}
```
davish, line 442: This should be an array, right? One per course instead of one per course code. Also will maintain the ordering. Maybe the format could be `[{course: string, response: {...}}]`

backend/alert/views.py
```diff
@@ -229,6 +235,249 @@ class RegistrationViewSet(AutoPrefetchViewSetMixin, viewsets.ModelViewSet):
...
+            res, normalized_course_code, reg = register_for_course(
+                course_code=course,
+                source="PCA",
+                user=request.user,
+                auto_resub=request.data.get("auto_resubscribe", False),
+                registration_group=registration_group,
+            )
+
+            # converting the codes to a proper response
+            response_codes[course] = self.response_code_convert(
```
davish, line 476: Change down here to a `response_codes.append(...)`

backend/alert/views.py
```diff
@@ -260,43 +509,21 @@ def handle_registration(request):
...
-                status=status.HTTP_500_INTERNAL_SERVER_ERROR,
-            )
+        obj = RegistrationViewSet.response_code_convert(
+            reg=reg,
+            normalized_course_code=normalized_course_code,
+            res=res,
+            section_code=section_code,
+        )
+
+        return_dict = {
```
davish, line 519: could this be flattened at all? the conditionals here make it kind of confusing to read.

davish (changes requested):
Added some comments, but great work here!

backend/alert/views.py
tanyew3st: Im not sure how to get around this because I want to extract the stuff in the dict to a response

PR #240, opened 2020-11-08
Transition miscellaneous files to TS (optional meetings and course ratings)

jeffrey-xiao-17:
Couple of notes:
1. DropdownButton.tsx -- the type of `children` props was Node, but `React.Children.map` doesn't seem to recognize it...  Added the any type for now
2. types.ts -- added an optional id field for MeetingBlock since `getConflictGroups` in `meetUtil.tsx` adds a temporary id field to each MeetingBlock


frontend/plan/components/DropdownButton.tsx
```diff
@@ -1,14 +1,27 @@
 import React, { useState } from "react";
-import PropTypes from "prop-types";
-import { useOnClickOutside } from "pcx-shared-components/src/useOnClickOutside";
+import { useOnClickOutside } from "../../shared-components/src/useOnClickOutside";
+
+interface DropdownButtonProps {
+    title: string;
+    children: Node;
+    filterData:
```
davish, line 7: Since this type annotation is copy/pasted, would be good to pull it out into a named type, like `FilterType` or something.

frontend/plan/components/meetUtil.tsx
```diff
@@ -38,14 +46,17 @@ export const getConflictGroups = (meetings) => {
...
+        conflicts[m2.id as number] = new Set([
+            ...Array.from(conflicts[m1.id as number]),
+            ...Array.from(conflicts[m2.id as number]),
+        ]);
+        conflicts[m1.id as number] = conflicts[m2.id as number];
     };
 
     meetings.forEach((m) => {
-        conflicts[m.id] = new Set([m]);
+        conflicts[m.id as number] = new Set([m]);
```
davish, line 59: wondering why we need to cast the `id` to be a `number` everywhere? looks like it's defined as a number in the types file. Is it because it may be undefined? If so, I know it's annoying, but to fully take advantage of the type system we should do an `if (m.id)` before doing anything with the `id` property.

frontend/plan/components/ShowMore.tsx
```diff
@@ -1,15 +1,22 @@
 import React, { useState } from "react";
-import PropTypes from "prop-types";
 import Truncate from "react-truncate-markup";
 
+interface ShowMoreProps {
+    more: Element | string;
```
davish, line 5: `ReactNode` might work here. Also, instead of declaring `children` as an `Element | string`, you can wrap this in `PropsWithChildren<ShowMoreProps>` in the type annotation. A bit cleaner!

davish (changes requested):
Some small comments, but this is looking good! Make sure to pull down before moving on, since I fixed some small merge conflicts.

frontend/plan/components/DropdownButton.tsx
```diff
@@ -67,7 +83,8 @@ export function DropdownButton({
                 <div className="dropdown-content">
                     {/* This injects the setIsActive method to allow children */}
                     {/* to change state of dropdown  */}
-                    {React.Children.map(children, (c) =>
+                    {/* TODO: replace "any" type here */}
+                    {React.Children.map(children, (c: any) =>
```
davish, line 87: I believe the type here is either `ReactNode` or `ReactElement`

frontend/plan/components/Cart.tsx
```diff
@@ -1,10 +1,23 @@
...
 import CartSection from "./CartSection";
 import { meetingsContainSection, meetingSetsIntersect } from "./meetUtil";
 import { removeCartItem, toggleCheck, fetchCourseDetails } from "../actions";
 
+import { ThunkDispatch } from "redux-thunk";
+import { Section, CartCourse } from "../types";
+
+interface CartProps {
+    courses: CartCourse[];
+    toggleCourse: (courseId: Section) => void;
```
davish, line 12: would rename this from `courseId` to `section` in the signature to be more specific. There's definitely a conflation between `course` and `section` in the code (we inherited that unfortunately), would be good to start moving things over to be more specific.

frontend/plan/components/Cart.tsx
```diff
@@ -1,10 +1,23 @@
...
+    removeItem: (courseId: string) => void;
```
davish, line 13: `sectionId` here would be good

frontend/plan/components/Cart.tsx
```diff
@@ -86,24 +99,12 @@ const Cart = ({
...
-
-// const mapStateToProps = ({ schedule: { cartSections, schedules, scheduleSelected } }) => ({
 const mapStateToProps = ({
     schedule: { cartSections = [], schedules, scheduleSelected, lastAdded },
     sections: { courseInfoLoading },
-}) => ({
+}: any) => ({
     courseInfoLoading,
-    courses: cartSections.map((course) => ({
+    courses: cartSections.map((course: any) => ({
```
davish, line 107: I know I said not to add any types in the redux things, but i think it would be useful to add some types on these new variables. For example, we know `course` is a `Section` (and would probably be good to change its name in the process)

frontend/plan/components/Cart.tsx
```diff
@@ -112,18 +113,18 @@ const mapStateToProps = ({
         overlaps: meetingSetsIntersect(
             course.meetings,
             schedules[scheduleSelected].meetings
-                .filter((s) => s.id !== course.id)
-                .map((s) => s.meetings)
+                .filter((s: any) => s.id !== course.id)
```
davish, line 116: Again here: `s: Section`

davish (approved):
Looks great! Some small comments but excited to merge this :D

PR #241, opened 2020-11-09
Improve error when registrar API request fails

mureytasroc:
Currently an iteration on None error will be raised no matter what the error returned from the Registrar is.


davish (approved):
Looks good!

PR #244, opened 2020-11-10
Test different semester same requirement

mureytasroc:


backend/tests/plan/test_api.py
```diff
@@ -126,6 +126,17 @@ def test_filter_for_req(self):
         self.assertEqual(1, len(response.data))
         self.assertEqual("MATH-114", response.data[0]["id"])
 
+    def test_filter_for_req_dif_sem(self):
+        req2 = Requirement(
+            semester=("2019A" if TEST_SEMESTER == "2019C" else "2019C"), code="REQ", school="SAS"
+        )
+        req2.save()
+        req2.courses.add(self.math)
```
davish, line 134: would be cool if this requirement also had a different course attached to it, so we could make sure only the current semester's requirement was being used in the query.

davish (approved):
Quick suggestion but g2g!

PR #245, opened 2020-11-24
Support asynchronous classes

davish:
Not a new JS feature, but support classes without meeting times in Penn Course Plan.


ArmaanT (approved):
I can't believe doing a null check is a bad thing

PR #246, opened 2020-11-27
Pca stats

mureytasroc:


backend/courses/models.py
```diff
@@ -399,6 +408,35 @@ class Meta:
...
+        NOTE: sections with an invalid class capacity (0 or negative) are excluded from
+        computation of the statistic, and if this section has an invalid class capacity, then
+        this property will equal None (or null in JSON).
+        """
+        from alert.models import Registration  # imported here to avoid circular imports
+
+        if self.capacity == 0:
+            return None
+        aggregate_scores = (
+            Registration.objects.filter(section__course__semester=get_current_semester(),
```
davish, line 427: Does this query need to be run every time? Not sure if it's an expensive one, but `aggregate_scores` is the same across all sections, so wondering if there's any way to cache it easily.

davish (changes requested):
All around looks great! Some nits, but also some questions/opinions of mine that I'm glad to talk over more.

backend/alert/serializers.py
```diff
@@ -148,3 +148,49 @@ class Meta:
...
+            """
+        ),
+    )
+
+    def get_course_title(self, obj):
+        return obj.course.title
+
+    class Meta:
+        model = Section
+        fields = [
```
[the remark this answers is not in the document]
mureytasroc: Yeah looking at the docs you are def right that read_only=True is redundant. I think you still need to define fields and read_only_fields separately in Meta though. I could be wrong, but the only place in `rest_framework/serializers.py` where read_only_fields is mentioned is in the following code (which does not seem to create fields in the same way that the fields list is used in `get_field_names`):

```
def get_extra_kwargs(self):
        """
        Return a dictionary mapping field names to a dictionary of
        additional keyword arguments.
        """
        extra_kwargs = copy.deepcopy(getattr(self.Meta, 'extra_kwargs', {}))

        read_only_fields = getattr(self.Meta, 'read_only_fields', None)
        if read_only_fields is not None:
            if not isinstance(read_only_fields, (list, tuple)):
                raise TypeError(
                    'The `read_only_fields` option must be a list or tuple. '
                    'Got %s.' % type(read_only_fields).__name__
                )
            for field_name in read_only_fields:
                kwargs = extra_kwargs.get(field_name, {})
                kwargs['read_only'] = True
                extra_kwargs[field_name] = kwargs

        else:
            # Guard against the possible misspelling `readonly_fields` (used
            # by the Django admin and others).
            assert not hasattr(self.Meta, 'readonly_fields'), (
                'Serializer `%s.%s` has field `readonly_fields`; '
                'the correct spelling for the option is `read_only_fields`.' %
                (self.__class__.__module__, self.__class__.__name__)
            )

        return extra_kwargs
```

backend/alert/urls.py
```diff
@@ -13,6 +14,11 @@
 
 urlpatterns = [
     path("courses/", courses.views.SectionList.as_view(), name="section-search"),
+    path(
+        "sections-with-statistics/",
```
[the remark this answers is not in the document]
mureytasroc: How about `section-stats`, since `sections/stats` might be weird for retrieve requests... it would be something like `/api/sections/stats/{id}/`?

backend/alert/serializers.py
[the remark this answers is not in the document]
davish: I could be wrong from experimentation but I think it's either-or: either define `read_only=True` on the field, or specify `read_only_fields`

backend/alert/serializers.py
```diff
@@ -5,6 +5,12 @@
 from alert.models import Registration
 from courses.models import Section, StatusUpdate, string_dict_to_html
 
+from django.db.models import Count, Max, Min
+
+from courses.util import get_current_semester
+
+from django.core.cache import cache
```
davish, line 12: Generally, all the `django` imports should go above all our first-party imports

backend/alert/serializers.py
```diff
@@ -148,3 +154,102 @@ class Meta:
...
+    """
+    This function normalizes the given value to a 0-1 scale based on the given min and max.
+    Raises ValueError if min >= max.
+    """
+    if min >= max:
+        raise ValueError(f"normalize called with min >= max ({min} >= {max})")
+    return float(value - min) / float(max - min)
+
+
+class SectionStatisticsSerializer(serializers.ModelSerializer):
```
davish, line 169: A docstring here explaining the section stats would be great :)

davish (approved):
Looks great to me! Really like the separation of concerns with the separate view/serializer. One or two nits but after that go ahead and merge away!

backend/alert/serializers.py
mureytasroc: Yeah this isn't linted yet.

backend/alert/serializers.py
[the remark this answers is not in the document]
mureytasroc: Yeah I removed all the "read_only=True"s

dlike230 (changes requested):
Overall, looks good! Just a few small things that I put inline comments for.
I wasn't able to test yet because of backend issues, but I'll try again soon

frontend/review/tsconfig.json
```diff
@@ -0,0 +1,25 @@
...
+    "module": "esnext",
+    "moduleResolution": "node",
+    "resolveJsonModule": true,
+    "isolatedModules": true,
+    "noEmit": true,
+    "jsx": "react"
+  },
+  "include": [
+    "src"
+  ]
```
sigmachirality, line 24: huge code quality moves <3

frontend/review/src/components/GraphBox.js
```diff
@@ -0,0 +1,501 @@
...
+    while (j >= 0 && seriesData[j][0] >= xVal - windowSize) {
+      total += seriesData[j][1];
+      numInTotal += 1;
+      j -= 1;
+    }
+    let movingAverageVal = total / numInTotal;
+    averageData.push({
+      x: (xVal * 100).toFixed(2),
+      y: movingAverageVal.toFixed(2),
+    });
```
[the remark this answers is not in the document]
sigmachirality: added benefit is that using .map will let you make these variables const. Actually, is there a reason why any of them need to be reassignable?

frontend/review/src/components/GraphBox.js
```diff
@@ -0,0 +1,501 @@
+import React, { useState, useEffect } from "react";
+import styled from "styled-components";
+import { defaults, Scatter } from "react-chartjs-2";
+
+let addDropDate = [];
```
sigmachirality, line 5: why is this not const?

frontend/review/src/components/GraphBox.js
```diff
@@ -0,0 +1,501 @@
...
+
+//PCA Demand Chart Data
+const genDemandChartData = (data, averageData) => {
+  return {
+    datasets: [
+      {
+        type: "line",
+        label: "5% Moving Average",
+        data: averageData,
+        borderColor: "#6378E9",
```
sigmachirality, line 98: someone should make a card to extract colors out into a colors constants file somewhere

frontend/review/src/components/GraphBox.js
```diff
@@ -0,0 +1,501 @@
...
+            data["datasets"][0]["data"][toolTipItem[0]["index"]].x / 100
+          ) +
+          "\n";
+        bodyText +=
+          "Registration Difficulty: " +
+          data["datasets"][0]["data"][toolTipItem[0]["index"]].y +
+          "\n";
+        bodyText +=
+          "5% Moving Average: " +
+          data["datasets"][1]["data"][toolTipItem[0]["index"]].y;
```
sigmachirality, line 176: would it be possible to use a js format string to generate this instead of string concats?

frontend/review/src/components/GraphBox.js
```diff
@@ -0,0 +1,501 @@
...
+          autoSkip: true,
+          beginAtZero: true,
+          maxTicksLimit: 6,
+          stepSize: 20,
+          maxRotation: 0,
+          minRotation: 0,
+          max: 100,
+          callback: function(value) {
+            return value + "%";
+          },
```
sigmachirality, line 213: prefer lambda function in this case

frontend/review/src/components/GraphBox.js
```diff
@@ -0,0 +1,501 @@
...
+            addDropDate.start,
+            addDropDate.end,
+            data["datasets"][0]["data"][toolTipItem[0]["index"]].x / 100
+          ) +
+          "\n";
+        bodyText +=
+          "% of Sections Open: " +
+          data["datasets"][0]["data"][toolTipItem[0]["index"]].y +
+          "%";
+        return bodyText;
```
sigmachirality, line 273: format string?

frontend/review/src/components/GraphBox.js
```diff
@@ -0,0 +1,501 @@
...
+  const season = semester.substring(4);
+  if (season === "A") {
+    return "Spring " + year;
+  } else if (season === "B") {
+    return "Summer " + year;
+  } else if (season === "C") {
+    return "Fall " + year;
+  }
+  return year;
+};
```
sigmachirality, line 349: check in the utils file - I think there's already a function that does this, or something very close you can take advantage of.

frontend/review/src/components/GraphBox.js
```diff
@@ -0,0 +1,501 @@
...
+
+const calcApproxDate = (startDateString, endDateString, percent) => {
+  let startDate = new Date(startDateString);
+  let endDate = new Date(endDateString);
+
+  let percentageThrough = (endDate.getTime() - startDate.getTime()) * percent;
+  let approxDate = new Date(startDate.getTime() + percentageThrough);
```
sigmachirality, line 356: why are these not const?

frontend/review/src/components/GraphBox.js
```diff
@@ -0,0 +1,501 @@
...
+  return approxDate.toLocaleDateString("en-US", {
+    month: "short",
+    day: "numeric",
+  });
+};
+
+const GraphBox = ({ courseCode, courseData, isAverage }) => {
+  const [loaded, setLoaded] = useState(false);
+  const [pcaDemandChartData, setPCADemandChartData] = useState();
+  const [percentSectionsChartData, setPercentSectionsChartData] = useState();
```
sigmachirality, line 367: adding some plausible default states here could save trouble down the line

frontend/review/src/components/DetailsBox.js
```diff
@@ -55,240 +55,256 @@ const formsCol = {
...
+                  __html:
+                    sectionsList
+                      .filter(
+                        ({ semester, comments }) =>
+                          semester === selectedSemester && comments
+                      )
+                      .map((info) => info.comments)
+                      .join(", ") ||
+                    "This instructor does not have any comments for this course.",
+                }}
```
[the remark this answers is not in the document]
sigmachirality: The comments have raw html in them, so we do this to format accordingly.

frontend/review/src/components/ScoreBox.js
```diff
@@ -392,8 +409,12 @@ class ScoreBox extends Component {
   }
 
   componentDidUpdate(prevProps) {
-    const { data, liveData } = this.props;
-    if (prevProps.data !== data || prevProps.liveData !== liveData) {
+    const { data, liveData, isCourseEval } = this.props;
+    if (
+      prevProps.data !== data ||
+      prevProps.liveData !== liveData ||
+      prevProps.isCourseEval !== isCourseEval
```
sigmachirality, line 416: would prefer destructuring prevProps instead of accessing it everywhere in this function

sigmachirality (changes requested):
🔥 🔥 🔥 🔥 🔥 

left some nits/concerns mostly regarding overuse of `let` instead of `const` and string concats over format strings

frontend/review/src/components/GraphBox.js
andyjiang3: I want to be able to access it outside of the functional component. After I get the data in the functional component, I want to change the value of it. Is there a better way of doing this?

andyjiang3: hmm I could just push it in

dlike230 (approved):

PR #247, opened 2020-11-29
Course recommendations prototype

dlike230:
Usage:
To train model: `python manage.py trainrecommender.py`
To test model with classes: `python manage.py recommendcourses --current_courses=CIS-110,CIS-160 --past_courses=ECON-001`
To test model with user: `python manage.py recommendcourses --user=dlike`


backend/courses/management/commands/trainrecommender.py
```diff
@@ -0,0 +1,278 @@
...
+    for course_index, cluster_index in enumerate(raw_cluster_result):
+        clusters[cluster_index].append(courses[course_index])
+
+    cluster_centroids = [sum(course_vectors_dict_curr[course] for course in cluster) / len(cluster) for cluster in
+                         clusters]
+    return cluster_centroids, clusters, course_vectors_dict_curr, course_vectors_dict_past
+
+
+def save_course_clusters(cluster_data):
+    return pickle.dump(cluster_data, open("./course-cluster-data.pkl", "wb"))
```
davish, line 269: We'll have to figure out a different way to do this in production, because the local filesystem isn't persistent -- any time we deploy new code, we're in a new docker container with a clean filesystem. We could dump the pickle into redis? That's longer-lived. @kirubarajan @pawalt would love your input here.

davish (commented):
I'm not super well-versed in sklearn so I don't think I'd be able to make a good review of that part of the code (which is most of it, haha). With regards to saving the results, though, we'll need to think of something better than pickle. It's awesome this is in a command so we can run it as a cron every morning. Redis seems to be the most naiive answer. We should set up a meeting with platform to go over ways to work with ML in kubernetes considering this is our first ML in production :D

backend/courses/management/commands/trainrecommender.py
pawalt: I'm partial to s3. A cronjob that dumps into s3 would make the most sense in my mind.

dlike230: I agree with @pawalt, s3 would be a pretty clean solution.

backend/courses/management/commands/trainrecommender.py
```diff
@@ -0,0 +1,312 @@
...
+
+    pickle.dump(cluster_data, open("./plan/temp_pickle/course-cluster-data.pkl", "wb"))
+
+    try:
+        S3.upload_file(
+            "./plan/temp_pickle/course-cluster-data.pkl", "penn.courses", "course-cluster-data.pkl"
+        )
+        print("success!")
+    except Exception:
+        print("there was a problem uploading the file / connecting to AWS!")
```
dlike230, line 304: It might be a good idea to print out the error that was caught too.

backend/plan/views.py
```diff
@@ -1,19 +1,93 @@
...
+
+        S3.download_file(
+            "penn.courses", "course-cluster-data.pkl", "./plan/temp_pickle/course-cluster-data.pkl"
+        )
+
+    return pickle.load(open("plan/temp_pickle/course-cluster-data.pkl", "rb"))
+
+
+@api_view(["POST"])
+@schema(PcxAutoSchema())
```
mureytasroc, line 43: Look at `PennCourses/docs_settings.py` for instructions on how to document this function-based view.

backend/courses/management/commands/trainrecommender.py
kirubarajan: I'm probably late here but S3 would be the industry standard for something like this.

backend/plan/views.py
```diff
@@ -1,19 +1,171 @@
...
+
+    if curr_courses or past_courses:
+        try:
+            user_vector, user_courses = vectorize_user_by_courses(
+                clean_course_input(curr_courses),
+                clean_course_input(past_courses),
+                curr_course_vectors_dict,
+                past_course_vectors_dict,
+            )
+        except Exception:
```
mureytasroc, line 142: Is there a way to catch a more specific exception or otherwise eliminate this "except Exception" catch-all solution (which could make it very hard to debug errors in the `vectorize_user_by_courses` function)?

backend/PennCourses/settings/base.py
```diff
@@ -181,3 +182,10 @@
 }
 
 STATS_WEBHOOK = os.environ.get("STATS_WEBHOOK", None)
+
+S3 = boto3.client(
+    "s3",
+    region_name="us-east-1",
+    aws_access_key_id=os.environ.get("AWS_ACCESS_KEY_ID"),
+    aws_secret_access_key=os.environ.get("AWS_SECRET_ACCESS_KEY"),
```
ArmaanT, line 190: What happens if you remove these lines? In the future, we're planning on moving away from providing explicit access/secret keys in production and I'm pretty sure boto3 will automatically find the credentials, but I'm not 100% sure

mureytasroc: Boto3 will look in several locations when searching for credentials. The mechanism in which Boto3 looks for credentials is to search through a list of possible locations and stop as soon as it finds credentials. The order in which Boto3 searches for credentials is:

1. Passing credentials as parameters in the boto.client() method
2. Passing credentials as parameters when creating a Session object
3. Environment variables
4. Shared credential file (~/.aws/credentials)
5. AWS config file (~/.aws/config)
6. Assume Role provider
7. Boto2 config file (/etc/boto.cfg and ~/.boto)
8. Instance metadata service on an Amazon EC2 instance that has an IAM role configured.

[https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html)

mureytasroc: So are we good to omit those kwargs / assume the creds will be found elsewhere? Also, maybe we want to move the boto3 client instantiation to production.py if we won't be using S3 in the code in dev at all?

ArmaanT: Yep, environment variables will be set in production. I think moving the config to production makes sense, but that would break the endpoints in dev. Would need to find a fix/workaround for it somehow. I also just realized that the load pcr data management command uses S3. Do we have two different sets of config for boot in this repo?

backend/courses/management/commands/trainrecommender.py
mureytasroc: I'm thinking S3 and Redis are the optimal solution here (check out the most recent version of the PR)

PR #249, opened 2020-12-11
hotfix for alert showing nonexistent classes that have status updates

davish:
A longer term solution would be adding a field on sections and courses that explicitly has the date when the course was last included in a registrar scrape. We'd then filter to make sure that field was not null, or sufficiently new, before displaying it to the user as a course. This would effectively hide *only* courses that were in status updates but not on PIT.


ArmaanT (approved):
LGTM! I like the long-term plan

PR #260, opened 2021-01-06
Update Documentation and Remove Current Semester Routes

mureytasroc:


backend/courses/views.py
```diff
@@ -171,9 +200,11 @@ class RequirementList(generics.ListAPIView, BaseCourseMixin):
     schema = PcxAutoSchema(
         examples=examples.RequirementList_examples,
         response_codes={
-            "/api/plan/requirements/": {"GET": {200: "[SCHEMA]Requirements listed successfully."}},
-            "/api/courses/{semester}/requirements/": {
-                "GET": {200: "[SCHEMA]Requirements listed successfully."}
+            "/api/plan/requirements/": {
+                "GET": {200: "[DESCRIBE_RESPONSE_SCHEMA]Requirements listed successfully."}
+            },
+            "/api/base/{semester}/requirements/": {
```
davish, line 206: This should probably use the `reverse()` function in django instead of hard-coding in URLs -- sorry for not catching this earlier, I know this isn't strictly new for this PR. I think as a general rule, URLs should always only show up in the URL file so we don't have to do big find-and-replaces if we ever change URLs (like now, haha)

frontend/plan/actions/index.js
```diff
@@ -172,7 +172,7 @@ export const clearSchedule = () => ({
 });
 
 export const loadRequirements = () => (dispatch) =>
-    doAPIRequest("/requirements/").then(
```
davish, line 175: do we never use `doAPIRequest` anymore? it also sets some headers/cookies that we need, so I think the answer is to modify `doAPIRequest` to not default to `/api/plan` but simply `/api`, rather than circumvent it altogether.

davish (changes requested):
Two comments -- will take another look later too!

frontend/plan/actions/index.js
dlike230: Yeah that makes sense

dlike230 (changes requested):
Should be good once Davis's suggestions about doApiRequest are implemented

backend/plan/views.py
```diff
@@ -9,37 +9,11 @@
...
-    """
-
-    schema = PcxAutoSchema(
-        examples=examples.CourseListSearch_examples,
-        response_codes={
-            "/api/plan/courses/": {"GET": {200: "[SCHEMA]Courses listed successfully."}}
-        },
-    )
-
-    filter_backends = [TypedCourseSearchBackend, CourseSearchFilterBackend]
```
davish, line 39: This might be further down, but does this get replicated anywhere? The PCP search endpoint does actually add filters on top of the base CourseList, so want to make sure we still have that.

backend/courses/views.py
```diff
@@ -16,8 +16,9 @@
     StatusUpdateSerializer,
     UserSerializer,
 )
-from PennCourses.docs_settings import PcxAutoSchema
-from plan.search import TypedSectionSearchBackend
+from PennCourses.docs_settings import PcxAutoSchema, reverse_func
+from plan.filters import CourseSearchFilterBackend
+from plan.search import TypedCourseSearchBackend, TypedSectionSearchBackend
```
davish, line 21: I don't think we want to have `courses` depending on `plan` and `plan` depending on `courses`, it's not a strict circular dependency issue, but more a question of separation of concerns... But on the other hand, we do have some mentions of `review` all over. Not sure what the common practice is, if we should care that our apps are referencing each other or not.

davish (approved):
This looks good to me! We should go through and make sure everything still works on PCP/PCA before merging, though. Had one comment about moving things from plan to courses, which we can talk about too.

backend/plan/views.py
mureytasroc: Yeah it is replicated here:

https://github.com/pennlabs/penn-courses/pull/260/files/e9edc68b35b84a5d2e8cbf0d135e00d69fae3c77#diff-52fd16120332ef10fba01c425d68eff2f29194a6a128ac1c231fb0f856deb6c5R132

The new url is `/api/base/<slug:semester>/search/courses/`

dlike230 (changes requested):
Searching up course details doesn't seem to work. I get this error:

```
Using the URLconf defined in PennCourses.urls, Django tried these URL patterns, in this order:
__debug__/
admin/doc/
admin/
accounts/me/ [name='user-view']
accounts/
api/ review/
api/ plan/
api/ alert/
api/ base/ <slug:semester>/courses/ [name='courses-list']
api/ base/ <slug:semester>/search/courses/ [name='courses-search']
api/ base/ <slug:semester>/courses/<slug:full_code>/ [name='courses-detail']
api/ base/ <slug:semester>/search/sections/ [name='section-search']
api/ base/ <slug:semester>/sections/<slug:full_code>/ [name='sections-detail']
api/ base/ <slug:semester>/requirements/ [name='requirements-list']
api/ base/ statusupdate/<slug:full_code>/ [name='statusupdate']
api/ options/
api/ openapi/ [name='openapi-schema']
api/ documentation/ [name='documentation']
webhook [name='webhook']
The current path, api/base/current/search/courses/CIS-105, didn't match any of these.
```

dlike230 (approved):
lgtm!

PR #263, opened 2021-01-12
Day and Time Filter for PCP

tanyew3st:
Added two filters in the backend - one for day and one for time. This will filter out a current query to make sure that a course is included if and only if it has a section where all the meetings are contained within the day/time parameters.


backend/courses/models.py
```diff
@@ -395,6 +391,27 @@ class Meta:
         db_index=True,
         help_text="The number of credits this section is worth.",
     )
+    meeting_days = models.CharField(
```
mureytasroc, line 394: Are we sure we want to create new fields for meeting_days, earliest_meeting, and latest_meeting? It seems like it would be more appropriate to write an efficient query to compute these data from the existing fields and cache the result:
https://docs.djangoproject.com/en/3.1/topics/cache/#the-low-level-cache-api

davish: I'm not sure caching makes sense here, considering there's 5 * 48 * 2 possible queries, how many hits each one would get is kind of in question. I'm all for hopping on a call and talking through the current solution, though, and seeing where we can go from here

davish: Sorry, there'd be even more queries than that, considering the time/day filters can be combined with any of the existing filters, including the search box. We should look into how cacheing can work with search filters more generally, but in my current understanding I'm not sure how well it can fit.

davish (changes requested):
Good work! Still some efficiency concerns on saving this data, and some comments on code organization in filters.py. But I like where we ended up!

backend/courses/models.py
mureytasroc: I'm not saying cache the responses for each possible combination of filters (I agree this is very untenable), rather cache the queries for meeting_days, earliest_meeting, and latest_meeting per course, rather than storing them as new fields in the model. For instance we could make those fields [cached_properties](https://docs.djangoproject.com/en/3.1/ref/utils/#module-django.utils.functional). So that should lead to at most `3 * number_of_courses` total caches right? 

The only reason I say this is because it seems these fields are really derivative information from `sections -> meeting_times`, so adding these new fields would introduce redundancies into editable fields.

Obviously at a certain point, if a query is too expensive, it becomes worth it to introduce redundant fields (although updatable redundant fields could be very problematic if one of the fields is updated and the other isn't due to developer error; https://www.softkraft.co/7-common-mistakes-that-django-developers-make/#redundant-model-fields). But I wonder if this is really such an expensive query that it is worth it to introduce redundancy?

Even with caching, redundancy is an issue (but much less of an issue); I wonder if it is possible to set up a hook that resets a cache on the derived properties if the source fields are changed.

mureytasroc: Oh nvm you can't filter based on a property. Lol ignore all of the above.

backend/courses/models.py
```diff
@@ -416,6 +433,19 @@ def is_open(self):
 
     def save(self, *args, **kwargs):
         self.full_code = f"{self.course.full_code}-{self.code}"
+
+        # adding in meeting days
+        day_list = [meeting.day for meeting in self.meetings.all()]
```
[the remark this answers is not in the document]
mureytasroc: Maybe by using [signals](https://stackoverflow.com/a/25503326/14365776) which re-compute the property anytime the meetings change. And then we could make the save function compute the fields if they are null (and make their default be null).

PR #264, opened 2021-01-14
Convert rest of PCP components to Typescript

davish:
This PR converts the last few component files from javascript to typescript.

Redux is still purposely left untouched, since the goal is to move from redux for most server-side data fetching to [swr](https://swr.vercel.app).


frontend/plan/pages/index.tsx
```diff
@@ -142,120 +142,154 @@ function Index() {
...
+                                >
+                                    <Selector mobileView={true} view={0} />
+                                </div>
+                            </div>
+                        </div>
+                        <div style={{ padding: "10px" }}>
+                            <Cart setTab={setTab} mobileView={true} />
+                        </div>
+                        <div style={{ padding: "10px" }}>
+                            {/* Unfortunately running into a weird issue with connected components and TS. */}
```
jxiao, line 216: Out of curiosity, what was the specific issue here?

jxiao (approved):
Looks good to me - just two small imports that we can remove!

frontend/plan/pages/index.tsx
davish: honestly, no clue. Was saying a prop that was defined on the interface, wasn't actually defined. think it has to do with `connect` but can't be 100% sure

davish:
Closing since this was merged into #270

PR #265, opened 2021-01-14
Improve PCA

mureytasroc:
- Change close notifications to be sent in separate email threads from normal alerts (so subject is not "DEPT-XXX-XXX is now open!" when the notification is actually about the course closing).
- Expose `last_notification_sent_at` property for frontend (and use in the manage alerts page), the most recent `notification_sent_at` value across all of a user's historical registrations for a given registration's section. I think this better fits a user's natural understanding of the "LAST NOTIFIED" column on the manage alerts page.
- Prevent resub when active registration already exists on backend (and prevent users from doing this on PCA by changing `registrations-list` to only include the most recently active registration for any given section, i.e. `registrations-list` will not contain multiple registrations from the same section).
- Prevent close notifications from being sent when a later registration in the resubscribe chain is canceled or deleted
- Add automatic inferring of property description from docstrings in the docs (request/response schemas)
- Add caching to the `get_current_semester` function, with invalidation on the post_save signal from the Option model if the value of the changed Option is "SEMESTER"


backend/alert/models.py
```diff
@@ -405,27 +409,75 @@ def is_active(self):
...
+        True if the registration is waiting to send a close notification to the user
+        once the section closes.  False otherwise.
         """
-        for k, v in self.is_waiting_for_close_filter().items():
-            if getattr(self, k) != v:
-                return False
-        return True
+
+        # WARNING: you should be frugal with your usage of this property, since it hits the DB
+        # each time it is called.
```
mureytasroc, line 461: Let me know if there is a (reasonably simple) way to implement the `is_waiting_for_close` property without hitting the db; the issue is I really prefer to have one source of truth for this property (which currently is `is_waiting_for_close_filter`), while still being able to filter Registrations by this property. Since I am using Q filters now I'm not sure how I can check an instance with the filters without hitting the db; there is probably a way but I'm guessing it would be very complicated and involve some heavy reliance on the details of Django's implementation of the Q function...

backend/alert/serializers.py
```diff
@@ -43,34 +44,10 @@ class RegistrationSerializer(serializers.ModelSerializer):
...
-    )
-    is_waiting_for_close = serializers.SerializerMethodField(
-        read_only=True,
-        help_text=dedent(
-            """
-        True if the registration is waiting to send a close notification to the user
-        once the section closes.  False otherwise.
-        """
-        ),
-    )
```
mureytasroc, line 63: We no longer need these SerializerMethodFields since the auto-docs should infer help text from property docstring now, pending the documentation-update branch being merged.

davish:
A comment before I do a full review of the PR, I feel like "Course update" isn't necessarily an improvement. When I use PCA with emails (not often anymore, but sometimes still) the first thing I look at is the subject, and I feel like we should have all the info in there. Instead of changing the subject to be generic, could we have it toggle based on open/closed notification? It should have that info in the right function since we need it to render different templates.

mureytasroc:
> A comment before I do a full review of the PR, I feel like "Course update" isn't necessarily an improvement. When I use PCA with emails (not often anymore, but sometimes still) the first thing I look at is the subject, and I feel like we should have all the info in there. Instead of changing the subject to be generic, could we have it toggle based on open/closed notification? It should have that info in the right function since we need it to render different templates.

Good point. If we changed the subject for close notifications, however, the emails would be in different threads. Is that worth it to avoid ppl getting an email with a subject about a class opening but content about it closing (without changing the subject to be more generic)?

davish:
Yeah, I think different threads is fine. The most important thing is people get the info as quickly as possible, which is mostly from the subject line.

mureytasroc:
Ok I changed it so that the subject like accurately communicates the message, but normal alerts aren't changed at all.

backend/alert/templates/alert/email_alert.html
```diff
@@ -6,7 +6,7 @@
         <img src="https://penncoursealert.com/assets/pca/PCA_logo.png" alt="Penn Course Alert Logo" style="width: 40px" />
     </a>
     <div style="text-align: center;">
-        <h1 style="margin-bottom: 1.5em;">{{ course }} just opened up!</h1>
+        <h1 style="margin-bottom: 1.5em;">{{ course }} just <span style="color:green;">opened up</span>!</h1>
```
davish, line 9: We may want to ask designers about what colors to use -- the default `red` and `green` in HTML can be pretty jarring.

backend/alert/templates/alert/email_close_alert.html
```diff
@@ -6,7 +6,7 @@
         <img src="https://penncoursealert.com/assets/pca/PCA_logo.png" alt="Penn Course Alert Logo" style="width: 40px" />
     </a>
     <div style="text-align: center;">
-        <h1 style="margin-bottom: 1.5em;">{{ course }} just closed.</h1>
+        <h1 style="margin-bottom: 1.5em;">{{ course }} just <span style="color:red;">closed</span>.</h1>
```
davish, line 9: ditto here

backend/alert/templates/alert/email_close_alert.html
```diff
@@ -6,7 +6,7 @@
         <img src="https://penncoursealert.com/assets/pca/PCA_logo.png" alt="Penn Course Alert Logo" style="width: 40px" />
```
davish, line 6: Sort of unrelated, but we may want to factor out the email template and use template inheritance so that we don't have to copy/paste between the two. Just a bit more DRY.

backend/alert/views.py
```diff
@@ -486,14 +517,23 @@ def get_queryset(self):
...
-        return Registration.objects.filter(
+        registrations = Registration.objects.filter(
             user=self.request.user,
             deleted=False,
             resubscribed_to__isnull=True,
             section__course__semester=get_current_semester(),
         )
+        # Now resolve conflicts where multiple registrations exist for the same section
+        # (by taking the registration with the later created_at date)
+        return registrations.filter(
```
davish, line 532: Is this query efficient? Worried about the subquery here, if it's being executed all in the DB or if we're getting a list back and then sending it back to the DB. Could you inspect with the debug toolbar?

davish (approved):
Some comments and questions, but generally looks good!

backend/alert/models.py
```diff
@@ -322,6 +323,101 @@ class Registration(models.Model):
...
+
+        return Registration.objects.filter(
+            *Registration.is_waiting_for_close_filter(), id=self.id
+        ).exists()
+
+    @property
+    def last_notification_sent_at(self):
```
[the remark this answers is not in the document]
mureytasroc: The issue is, this is actually meant to avoid looking at the last in chain, because the last in chain usually hasn't sent a notification yet -- so for instance if someone has auto resubscribe on, `notification_sent_at` of the last in chain would always be null -- and this is what happens on the frontend (a registration shows no alerts sent yet, even if alerts have indeed been sent for that section). Also we have to think beyond just resubscribe chains because someone could create a new registration instead of resubscribing and they would still expect "last notification sent at" to be the last time a notification was sent about that section. But if the efficiency of constricting the search to resubscribe chains would be worth it to avoid that edge case, then maybe we can look at the most recent registration in the chain which has a non-null `notification_sent_at` value

backend/PennCourses/docs_settings.py
```diff
@@ -1060,6 +1060,34 @@ def get_description(self, path, method):
             desc += '\n\n<span style="color:red;">User authentication required</span>.'
         return desc
 
+    # Overrides, uses overridden method
+    def map_serializer(self, serializer):
+        """
+        This method adds property docstrings as field descriptions when appropriate, in addition
+        to calling the overridden map_serializer function.
```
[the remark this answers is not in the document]
mureytasroc: Added a link to the docs -- I think adding a link to the line in the GitHub code would be bad practice since it will become incorrect if any changes are made to that document

backend/alert/models.py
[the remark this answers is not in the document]
davish: Ah I totally see, that makes a lot of sense. In that case though could you add a docstring explaining exactly what this is used for on the frontend? Would clear things up

backend/alert/views.py
mureytasroc: Yeah looks like it's all one query (since the queryset is evaluated lazily)

```
SELECT "alert_registration"."id",
       "alert_registration"."created_at",
       "alert_registration"."original_created_at",
       "alert_registration"."updated_at",
       "alert_registration"."source",
       "alert_registration"."api_key_id",
       "alert_registration"."user_id",
       "alert_registration"."email",
       "alert_registration"."phone",
       "alert_registration"."section_id",
       "alert_registration"."cancelled",
       "alert_registration"."cancelled_at",
       "alert_registration"."deleted",
       "alert_registration"."deleted_at",
       "alert_registration"."auto_resubscribe",
       "alert_registration"."notification_sent",
       "alert_registration"."notification_sent_at",
       "alert_registration"."close_notification",
       "alert_registration"."close_notification_sent",
       "alert_registration"."close_notification_sent_at",
       "alert_registration"."notification_sent_by",
       "alert_registration"."close_notification_sent_by",
       "alert_registration"."resubscribed_from_id"
  FROM "alert_registration"
  LEFT OUTER JOIN "alert_registration" T2
    ON ("alert_registration"."id" = T2."resubscribed_from_id")
 INNER JOIN "courses_section"
    ON ("alert_registration"."section_id" = "courses_section"."id")
 INNER JOIN "courses_course"
    ON ("courses_section"."course_id" = "courses_course"."id")
 WHERE (NOT "alert_registration"."deleted" AND T2."id" IS NULL AND "courses_course"."semester" = '''2020C''' AND "alert_registration"."user_id" = '1' AND "alert_registration"."created_at" IN (SELECT MAX(U0."created_at") AS "max_created_at" FROM "alert_registration" U0 LEFT OUTER JOIN "alert_registration" U1 ON (U0."id" = U1."resubscribed_from_id") INNER JOIN "courses_section" U2 ON (U0."section_id" = U2."id") INNER JOIN "courses_course" U3 ON (U2."course_id" = U3."id") WHERE (NOT U0."deleted" AND U1."id" IS NULL AND U3."semester" = '''2020C''' AND U0."user_id" = '1') GROUP BY U0."section_id"))
```

backend/alert/models.py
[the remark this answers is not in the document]
mureytasroc: Yeah will do

backend/alert/alerts.py
```diff
@@ -70,9 +70,37 @@ def send_alert(self, close_notification=False):
...
-        super().__init__(
-            "alert/email_alert.html", reg, close_template="alert/email_close_alert.html"
+        t = loader.get_template("alert/email_alert.html")
+        self.text = t.render(
+            {
+                "course": reg.section.full_code,
+                "brand": "Penn Course Alert",
+                "auto_resubscribe": reg.auto_resubscribe,
+                "course_event": '<span style="color:#78D381;">opened up</span>!',
+                "email_body": (
```
davish, line 80: I think `email_body` and `course_event` should still be in a template, I was more thinking about creating new templates that inherit using the [`extends` keyword](https://docs.djangoproject.com/en/3.1/ref/templates/language/)

PR #270, opened 2021-01-31
Schedule styled components

jxiao:
Removed CSS files:
- `slider.css`
- `schedule.css`
- `dropdown.css`

Moved all of schedule, most of selector (excluding `Selector.tsx`), and most of "search" to styled components (exception is `SearchBar.tsx` - about half transitioned because of bulma :/)


frontend/plan/components/search/SearchSortDropdown.tsx
```diff
@@ -25,22 +52,78 @@ const DropdownButton = ({
...
-    text: PropTypes.string,
-    onClick: PropTypes.func,
-    makeActive: PropTypes.func,
-};
-
 type SortByType = "Name" | "Quality" | "Difficulty" | "Good & Easy";
 const contents: SortByType[] = ["Name", "Quality", "Difficulty", "Good & Easy"];
 
+// Bulma: icon is-small
+const Icon = styled.span`
```
davish, line 65: ditto on this icon -- seems like this component is defined in a few places.

frontend/plan/components/selector/Selector.tsx
```diff
@@ -47,27 +62,10 @@ const Selector: FunctionComponent<SelectorProps> = ({
...
-            className="button is-loading"
-            style={{
-                height: "100%",
-                width: "100%",
-                border: "none",
-                fontSize: "3rem",
-            }}
-        />
-    );
+    const loadingIndicator = <Loading className="button is-loading" />;
```
davish, line 65: now that it's a one-line component, could it be inlined? or is `loadingIndicator` mentioned in multiple places?

I also wonder if the `is-loading` class could be turned into a styled component so we don't have to rely on stylesheets anymore. Maybe factor it out into a common file?

frontend/plan/components/selector/Tag.tsx
```diff
@@ -1,24 +1,49 @@
...
+    font-size: 0.75rem;
+    height: 2em;
+    justify-content: center;
+    line-height: 1.5;
+    padding-left: 0.75em;
+    padding-right: 0.75em;
+    white-space: nowrap;
+
+    &:hover {
+        background: ${({ isAdder }: { isAdder: boolean }) =>
```
davish, line 29: maybe a comment on what `isAdder` does?

frontend/plan/components/selector/Tag.tsx
```diff
@@ -1,24 +1,49 @@
 import React, { PropsWithChildren } from "react";
+import styled from "styled-components";
 
 interface TagProps {
     onClick?: () => void;
     isAdder?: boolean;
 }
 
+const TagContainer = styled.span`
```
davish, line 9:
```suggestion
const TagContainer = styled.span<{isAdder: boolean}>`
```

You can specify the props interface in the definition so the lambdas further down can infer the props. Also it'll help with inference when you use the container. I might have missed this in other places, but you can do it everywhere you have custom props in a styled component.

frontend/plan/components/selector/TagList.tsx
```diff
@@ -1,11 +1,22 @@
...
 interface TagListProps {
     elements: string[];
     limit?: number;
     onClick?: (id: string) => void;
 }
+
+const HiddenTagListContainer = styled.span`
+    overflow: hidden;
+    display: inline-block;
+    height: ${({ expanded }: { expanded: boolean }) => (expanded ? "auto" : 0)};
```
davish, line 14: Ditto on the interface here. Relevant docs: https://styled-components.com/docs/api#using-custom-props

frontend/plan/package-lock.json
```diff
@@ -0,0 +1,475 @@
+{
```
davish, line 1: we're using `yarn`, so make sure to delete this before merge.

davish (approved):
Really great work here! Some comments here and there, one about interfaces with typescript and styled components that might require a few changes.

I wanted to ask about `App.css`. doesn't look like you've removed any CSS rules there, although I'm sure there's a decent number that are no longer in use. I was wondering if you were planning on pruning that file in this PR or splitting that out into a separate one?

frontend/plan/components/search/CheckboxFilter.tsx
```diff
@@ -74,16 +103,13 @@ export function CheckboxFilter<
                                     }}
                                 />
                                 {/* eslint-disable-next-line jsx-a11y/label-has-for */}
-                                <label
-                                    htmlFor={filterProperty}
-                                    style={{ display: "table-cell" }}
-                                >
+                                <CheckboxLabel htmlFor={filterProperty}>
```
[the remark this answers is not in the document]
jxiao: Yep - looks like we still need this accessibility rule.  I'm looking into it, but the filters like CU and course type don't register without this attribute.

frontend/plan/components/selector/Selector.tsx
jxiao: `loadingIndicator` is in 2 places, but it's def able to be put in line - would probably be cleaner this way

jxiao: `is-loading` has been pulled out - is this used in several places?  Right now, I just have it in the `Selector.tsx` file, but if it's used across components, I could also pull it into the derived_components file

jxiao:
> I wanted to ask about `App.css`. doesn't look like you've removed any CSS rules there, although I'm sure there's a decent number that are no longer in use. I was wondering if you were planning on pruning that file in this PR or splitting that out into a separate one?

I haven't removed the CSS rules from there - Was planning on removing the individual CSS files first, then focusing on `App.css` since it's heftier than the rest

frontend/plan/components/selector/Selector.tsx
davish: This seems like a useful thing to have in the derived_components even if it's only in one place right now

jxiao: Done!

dlike230:
Looks good to me!

davish:
@jxiao merge whenever! Congrats on the two approvals :D

PR #276, opened 2021-03-05
Display recruiting banner on pca conditionally

dlike230:


frontend/alert/pages/index.tsx
```diff
@@ -132,6 +132,9 @@ function App() {
     useEffect(() => {
         ReactGA.initialize("UA-21029575-12");
         ReactGA.pageview(window.location.pathname + window.location.search);
+        fetch("/api/options/")
```
ArmaanT, line 135: Could we switch this to [platform's URL](https://platform.pennlabs.org/options/)? That way we can toggle all of these in one place

dlike230: Using that URL results in CORS issues. Is CORS enabled for that URL?

ArmaanT: Ah, that's my bad. I'll push a fix for that in the next day or so

PR #278, opened 2021-03-16
Update PCP schedule grid to 15min intervals

jxiao:
Incorporate 15 minute grid intervals to account for Penn's new block scheduling


davish (approved):
lets ride

PR #280, opened 2021-03-28
Course Recommendation Frontend

andyjiang3:
• Added course rec frontend

Currently using search result for course recommendations
Need to link with the backend when it is finished

Screen Shot 2021-03-28 at 1 05 13 PM
Screen Shot 2021-03-28 at 1 04 55 PM


frontend/plan/components/selector/Selector.tsx
```diff
@@ -58,20 +66,79 @@ const Selector: FunctionComponent<SelectorProps> = ({
...
 
+    const [showRecs, setShowRecs] = useState(true);
+    const [recCourses, setRecCourses] = useState([]);
+    const [recLoaded, setRecLoaded] = useState(false);
+    const [removeRec, setRemoveRec] = useState(true);
+
+    //delete func - does nothing for now
+    const onClickDelete = () => {};
+
+    useEffect(() => {
```
dlike230, line 77: This useEffect is missing some deps (such as setRecCourses)

frontend/plan/components/selector/Selector.tsx
```diff
@@ -58,20 +66,79 @@ const Selector: FunctionComponent<SelectorProps> = ({
...
+            headers: {
+                Accept: "application/json",
+                "Content-Type": "application/json",
+                "X-CSRFToken": getCsrf(),
+            },
+            body: JSON.stringify({}),
+        };
+
+        //@ts-ignore Not sure why this is not typechecking
+        fetch(`/api/plan/recommendations/`, requestOptions).then((res) =>
```
dlike230, line 95: We have a helper function for this, I believe it's called doApiRequest

dlike230 (changes requested):
Overall, great work! Only thing (other than my inline comments) is I'm not sure how we should go about updating the recommendations displayed (if at all) when a user's schedules change. Right now, if a user adds new classes that would affect their recommendations, that would have no effect on the recommendations displayed until the user refreshed the page. We could update recommndations every time the schedule is updated, but that might result in too many API calls. I'm tagging @mureytasroc and @davish to get their input.

mureytasroc:
> Overall, great work! Only thing (other than my inline comments) is I'm not sure how we should go about updating the recommendations displayed (if at all) when a user's schedules change. Right now, if a user adds new classes that would affect their recommendations, that would have no effect on the recommendations displayed until the user refreshed the page. We could update recommndations every time the schedule is updated, but that might result in too many API calls. I'm tagging @mureytasroc and @davish to get their input.
> 
> **Edit:** We should probably also remove the "x" buttons next to recommendations, since they don't do anything.

What do you think about adding a refresh button to "refresh recommendations"?

dlike230:
> > Overall, great work! Only thing (other than my inline comments) is I'm not sure how we should go about updating the recommendations displayed (if at all) when a user's schedules change. Right now, if a user adds new classes that would affect their recommendations, that would have no effect on the recommendations displayed until the user refreshed the page. We could update recommndations every time the schedule is updated, but that might result in too many API calls. I'm tagging @mureytasroc and @davish to get their input.
> > **Edit:** We should probably also remove the "x" buttons next to recommendations, since they don't do anything.
> 
> What do you think about adding a refresh button to "refresh recommendations"?

Seems like the best solution. Maybe for now we can just have recs update at a scheduled interval if changes have been made to the schedule?

andyjiang3:
• Added star next to rec courses in the search list
• Added error messages
• Added refresh button 

Screen Shot 2021-04-01 at 7 53 38 PM

frontend/plan/components/recomendations/RecBanner.tsx
```diff
@@ -15,35 +18,65 @@ const BannerContainer = styled.div<{ collapse: boolean }>`
...
+
 interface RecBannerProps {
     show: boolean;
     setShow: (_: boolean) => void;
+    setRefresh: React.Dispatch<React.SetStateAction<boolean>>;
 }
 
-const RecBanner = ({ show, setShow }: RecBannerProps) => {
+const RecBanner = ({ show, setShow, setRefresh }: RecBannerProps) => {
+    const [disabled, setDisabled] = useState(false);
```
dlike230, line 56: For the sake of clarity, let's call this `refreshDisabled`

frontend/plan/components/recomendations/RecInfo.tsx
```diff
@@ -45,17 +38,30 @@ const RecInfo = () => {
...
-                    Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed
-                    do eiusmod tempor incididunt ut labore et dolore magna
-                    aliqua. Ut enim ad minim veniam
+                    Recommendations are generated based on course descriptions
+                    and courses taken by users with similar mock schedules
+                    (voluntarily sent to Penn Labs for use in course
+                    recommendations by users of Penn Mobile). Refreshing the
+                    page or pressing the refresh button above the
+                    recommendations section will take into account any changes
+                    you have made to your PCP schedules since the last refresh.
```
dlike230, line 47: Replace PCP with Penn Course Plan

dlike230 (changes requested):
Just a couple really small things

dlike230 (approved):
LGTM!

PR #290, opened 2021-08-23
Fix unwanted behavior of toggling cart item when clicking on trash/info

eyingxuan:
Previously, there was a hack that prevents the cart item to be toggled when the trash/info icons are clicked, but styled-components generates unstable class names, so the hack only works in dev. 

Opted for using `event.stopPropagation()` which is nicer anyways.


jxiao (approved):
Thanks- looks good to me

PR #309, opened 2021-10-04
PCA head_registration field + id isolation, PCx section search course code

mureytasroc:

- Adds a `head_registration` field to the `Registration` model for efficient computation of resubscribe groups (as well as original registration and head registration) and `is_waiting_for_close` property. Adds a `compute_head_registrations` script to compute this field for all existing registrations. Adds code to the `resubscribe` method of `Registration` and the `loadregistrations_pca` script to automatically maintain the `head_registration` field (so the `compute_head_registrations` script should only be needed once). Removes the complex CTE sql queries for traversing resubscribe groups from `Registration`, as their functionality is easily replicated using the `head_registration` field.
- Removes the database call from the `is_waiting_for_close` property and changes `is_waiting_for_close_filter` to return a dictionary instead of a Q object.
- Adds `close_notification` properly to the PCA registration serializers/documentation.
- Catches cases of PCA registration id conflicting with a different user's registration id on create, and returns an informative error message.
- Adds a `course_code` field to the response of PCx Section Search (useful for PCA bulk registration frontend).


backend/alert/management/commands/compute_head_registrations.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def handle(self, *args, **kwargs):
+        print("Recomputing head registrations...")
+        queryset = Registration.objects.order_by("pk")
+        pk = -1
+        last_pk = queryset.order_by("-pk")[0].pk
+        pbar = tqdm(total=last_pk)
+        while pk < last_pk:
+            to_save = []
+            for registration in queryset.filter(pk__gt=pk)[:1000]:
```
AaDalal, line 29: Why only collect the first 1000 items out of the queryset? Are those the only ones lacking a head_registration field?

backend/alert/management/commands/compute_head_registrations.py
```diff
@@ -0,0 +1,40 @@
...
+                if pk >= 0:
+                    pbar.update(registration.pk - pk)
+                pk = registration.pk
+                head_registration = registration.get_most_current_iter()
```
AaDalal, line 33: where does `get_most_current_iter` come from?

AaDalal: Makes sense, thank you.

backend/alert/views.py
```diff
@@ -507,7 +512,7 @@ def update(self, request, pk=None):
         return Response({"detail": "no changes made"}, status=status.HTTP_200_OK)
 
     def create(self, request, *args, **kwargs):
-        if self.get_queryset().filter(id=request.data.get("id")).exists():
+        if Registration.objects.filter(id=request.data.get("id")).exists():
             return self.update(request, request.data.get("id"))
```
AaDalal, line 516: Consider if we find that an object has `id = request.data.get("id")` but that object is not `self`. `self` will be updated to have an `id` of `request.data.get("id")`, but what will the object that used to have an `id` of `request.data.get('id')` have?

backend/courses/util.py
```diff
@@ -459,3 +459,17 @@ def subquery_count_distinct(subquery, column):
         ),
         0,
     )
+
+
+def does_object_pass_filter(obj, filter):
+    """
+    Returns True iff the given obj satisfies the given filter dictionary.
```
AaDalal, line 466: Would it be helpful to add a note specifying that this really only works when the filter does not involve something like `gt`? If you included a filter like `user__id__gt` I believe this function would try to `getattr(actual_value, 'gt')` which would fail.

AaDalal (commented):
With the exception of a few questions [here](https://github.com/pennlabs/penn-courses/pull/309/files#r722793253) and [here](https://github.com/pennlabs/penn-courses/pull/309/files#r722797218), the changes make sense to me.

backend/courses/util.py
mureytasroc: Yeah very good point. Updated the docstring to reflect this (and also the fact that it can only traverse a relation to a single related object, not a related set).

backend/alert/views.py
mureytasroc: There's actually a check in the [update method](https://github.com/pennlabs/penn-courses/pull/309/files#diff-30321778847dd168745458b161cae6aa7914fecdc746484bb7267d2da2bdafd7R368) to throw a 404 (not found) response if an object with the given id doesn't exist in all `Regstration.objects`, and otherwise a 403 (forbidden) response if an object with the given id doesn't exist in the set of registrations that the user owns (the `queryset`), meaning a registration with the specified id exists but it isn't owned by the user who is requesting to modify it. 

This redirection to `self.update(request, request.data.get("id"))` if an object with the given id exists in all `Registration.objects` will cause an update if the user owns the registration (as the docs say it should), but if the user doesn't own the registration with the matching ID, it will return a 403. So there's no case where a registration id will actually be changed to an id that is already taken by a different registration (in which Django would throw an `IntegrityError` saying "UNIQUE constraint failed".

PR #312, opened 2021-10-22
PCP sort by recommendation

mureytasroc:

- Adds `recommendation_score` field to each course in the response of course search (`CourseListSearch` viewset). `recommendation_score` is None when the user is not logged, `self.request` is not defined in the viewset, or a particular course is not part of the recommendation model.
- Adds corresponding passing tests for when the user is logged in, not logged in, and when the user is logged in but the course is not part of the model.
- Replace `courses.views.CourseListSearch.get_serializer_context` with `courses.views.CourseList.get_serializer_context` by default for all tests. The code to do this is in `tests/__init__.py`. Note that this is default is only currently overriden in `backend.tests.courses.test_api.CourseSearchRecommendationScoreTestCase.test_recommendation_is_number_when_user_is_logged_in`
- Changes `trainrecommender` to allow for descriptions to be passed in as a CSV with rows of the form (`full_code`, `description`) using the `preloaded-descriptions-path` argument. Descriptions will be preferentially pulled from CSV when it is provided; if a course is not in the CSV then it is pulled from the database (if it does not exist in the DB, an empty string is used).  Previously, course descriptions were always queried from the database, which lead to problems when testing with incomplete databases.
- Changed the behavior of `retrieve_course_clusters` in `recommendcourses.py` so that, when provided by `, when in development mode, trains the model using `.../tests/plan/course_recs_test_data/course_descriptions_test.csv` for descriptions.
- (minor) Mock `retrieve_course_clusters` in `backend.tests.courses.test_api.CourseSearchRecommendationScoreTestCase.test_recommendation_is_number_when_user_is_logged_in` 
- (minor) adds the `cosine_similarity` helper function to `recommendcourses.py`


backend/plan/management/commands/trainrecommender.py
```diff
@@ -338,6 +346,15 @@ def train_recommender(
         ), "If you are training on data from S3, there's no need to supply a local data path"
     if course_data_path is not None:
         assert course_data_path.endswith(".csv"), "Local data path must be .csv"
+    if course_descriptions_path is not None:
+        assert course_descriptions_path.endswith(
+            ".csv"
+        ), "Local course descriptions path must be .csv"
+        assert course_data_path is not None, (
+            "If course_description_path is provided then course_data_path should also"
+            "be provided. Note that the courses should be ordered the same in"
```
mureytasroc, line 355: Looking at `generate_course_vectors_dict`, it seems that the courses don't actually have to be ordered the same in both csvs (the courses in `course_description_path` just have to be a superset of the `courses` in course_data)

backend/plan/management/commands/trainrecommender.py
```diff
@@ -374,7 +391,21 @@ def train_recommender(
             if course_data_path is not None
             else courses_data_from_db()
         )
-    course_clusters = generate_course_clusters(courses_data, n_per_cluster)
+
+    if course_descriptions_path is not None:
+        course_descriptions_or_none = courses_data_from_csv(course_descriptions_path)
+    else:
+        course_descriptions_or_none = None
```
mureytasroc, line 398: Could the default here be `dict()`, and if `course_descriptions_path` is not None then set it to `dict(courses_data_from_csv(course_descriptions_path))`? Then we could call this `preloaded_descriptions` or something like that, and in `generate_course_vectors_dict`, we could default to using `get_description` if the course code isn't in the `preloaded_descriptions` dict. Just a suggestion in case we want to relax the requirement that the courses in `course_description_path` have to be a superset of the courses in course_data.

backend/tests/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+        ```
+        @unittest.mock.patch('courses.views.CourseListSearch.get_serializer_context',
+               new=production_CourseListSearch_get_serializer_context)
+        ```
+    (Note that you should put
+    `from tests import production_CourseListSearch_get_serializer_context`
+    at the top of the file to make production_CourseListSearch_get_serializer_context
+    is defined. Note also that reverting may lead to a `botocore`
+    error in the CI tests. To resolve this `botocore` error it is suggested that you patch
+    `plan.management.commands.recommendcourses.retrieve_course_clusters`. See
```
mureytasroc, line 25: This should be `courses.views.retrieve_course_clusters` (as discussed on Slack), right?

backend/tests/courses/test_api.py
```diff
@@ -198,6 +202,77 @@ def test_search_by_instructor(self):
...
+        self.client.login(username=self.username, password=self.password)
+
+        self.course, self.section = create_mock_data("PSCI-437-001", TEST_SEMESTER)
+        response = self.client.get(
+            reverse("courses-search", args=["current"]), {"search": "PSCI-437", "type": "auto"}
+        )
+
+        self.assertEqual(200, response.status_code)
+        self.assertEqual(len(response.data), 1)
+        self.assertIs(response.data[0]["recommendation_score"], None)
```
mureytasroc, line 229: Very minor nit, but this could be replaced with `assertIsNone`

backend/tests/courses/test_api.py
```diff
@@ -198,6 +202,77 @@ def test_search_by_instructor(self):
...
+        self.client.login(username=self.username, password=self.password)
+
+        curr_semester_schedule = Schedule.objects.create(
+            person=self.user, name="curr_semester_schedule", semester=TEST_SEMESTER
+        )
+        curr_semester_schedule.save()
+        # NOTE: many of the sections in this schedule match up with the `semester`
+        # used to create the schedule
+        curr_semester_schedule.sections.add(
+            Course.objects.get(full_code="PSCI-498", semester="2019A").sections.get(code="001")
```
mureytasroc, line 256: Shouldn't the semester of each course in the schedule match the semester of the schedule? Not sure I understand the comment/note above.

backend/tests/plan/test_course_recs.py
```diff
@@ -48,6 +48,35 @@ def set_semester():
...
+        `/tests/plan/course_recs_test_data/course_data_test.csv`
+        and `/tests/plan/course_recs_test_data/course_descriptions_test.csv`.
+
+        The contents of `/tests/plan/course_recs_test_data/course_data_test.csv` are 3 columns:
+        - a `person_id` (used when creating schedules)
+        - a course `full_code` column (ie "PSCI-498")
+        - a semester column (ranging between 2016C and 2020A).
+
+        Courses are created with approximately the following specification:
+        - `department_id`: Corresponds to the department code embedded in the `code`
```
mureytasroc, line 62: ...embedded in the `full_code`?

backend/tests/plan/test_course_recs.py
```diff
@@ -48,6 +48,35 @@ def set_semester():
...
+           for a course is not "2020A" or "2017A" and the course `full_code` is not "HIST-650"
+           another course entry is created with `semester` equal to `TEST_SEMESTER` as defined
+           at the top of this file (2021C at the time of writing.)
+        - `description` : corresponds to the entry in
+           `/tests/plan/course_recs_test_data/course_descriptions_test.csv`
+
+        Sections corresponding to each created course are created with approximately this
+        specification
+        - `code` : "001"
+        - `full_code` : the course's `full_code` + "001"
```
mureytasroc, line 77: `+ "-001"`?

mureytasroc (commented):
Really awesome work here. I love all the edge-case checking, input validation, and documentation. Just a couple bugs in comments, some minor nits, and some suggestions about how we could relax input requirements.

backend/plan/management/commands/trainrecommender.py
AaDalal: Good point. That's some old documentation from a previous implementation that I changed. I'll remove that.

backend/tests/__init__.py
AaDalal: Yes, good catch

backend/tests/plan/test_course_recs.py
AaDalal: Good point

backend/tests/courses/test_api.py
AaDalal: That comment should read
> NOTE: many of the sections in this schedule **do not** match up with the `semester` **of the schedule**

I added it because we generally try to maintain the invariant that the schedule semester matches the semester of all the courses in the schedule. However `TEST_SEMESTER`, which is the variable that we used for the schedule's semester, does not necessarily match up with the course semesters that we place in the schedule.

mureytasroc: Ok sounds good

backend/plan/management/commands/trainrecommender.py
AaDalal: Yeah that makes a lot of sense.

The only issue I foresee is that now we're allowing mixing of preloaded descriptions (from the CSV) and database descriptions. This means we remove the invariant that `courses_data` and `course_descriptions` are coupled (ie course descriptions requires course_data, and the courses in the CSVs are expected to match up). I think most of the time when you provide both CSVs you would want them to be coupled, so this opens up some potentially confusing behavior.

With that being said, I think that we can bypass any confusion by providing good documentation.

mureytasroc: That's fair. If we do implement this, probably fine to leave it undocumented and remove the documentation describing the need for the courses in `course_descriptions` to be a superset of the courses in `courses_data` (maybe just say "descriptions will be taken from `course_descriptions` if specified, and any missing descriptions will be taken from the database, or set to the empty string if not found"). Generally the less need for documentation / explanation the better. Otherwise it's probably fine to keep it the way it is right now, your call.

backend/plan/management/commands/trainrecommender.py
```diff
@@ -392,19 +388,18 @@ def train_recommender(
...
-    else:
-        course_descriptions_or_none = None
+    preloaded_descriptions = dict()
+    if preloaded_descriptions_path is not None:
+        preloaded_descriptions = dict(courses_data_from_csv(preloaded_descriptions_path))
 
-    if course_descriptions_path is None and verbose:
+    if preloaded_descriptions_path is None and verbose:
         print(
             "A course_description_path has not been supplied."
```
mureytasroc, line 397: Might want to global search for `course_description_path` and replace with `preloaded_descriptions_path` just to stay consistent.

backend/plan/management/commands/trainrecommender.py
```diff
@@ -242,11 +241,17 @@ def generate_course_vectors_dict(courses_data, use_descriptions=True):
...
     copresence_vectors_by_course_past = vectorize_by_copresence(grouped_courses, as_past_class=True)
     courses_by_user = get_unsequenced_courses_by_user(grouped_courses)
-
     courses, courses_vectorized_by_schedule_presence = zip(
         *vectorize_courses_by_schedule_presence(courses_by_user).items()
     )
-    courses_vectorized_by_description = vectorize_courses_by_description(courses)
+
+    descriptions = []
+    for course in courses:
```
AaDalal, line 249: NOTE: This logic is not fully covered in the test cases because all the descriptions are pulled from the CSV so the `else` should never fire. However, I conducted some informal testing where I removed a course description from the csv and added it to the db (so that the `else` fired), and all worked.

AaDalal (approved):

PR #314, opened 2021-10-25
Day/Time Filter for PCP

jxiao:
Default Filter             |  Sample Filter
:-------------------------:|:-------------------------:
default  |  sample

**Done**
- Added day/time to filter bar as dropdown

**TODO**
- Fix overlap of filter bar with "Clear" button (issue is that adding an overflow scroll disables the dropdowns), only major issue on < 13in computers


dlike230 (changes requested):
Looks good! Just a couple small things
- We should probably narrow the time range filter, even if it doesn't technically cover the whole day (we can round off the ends of the filter to the actual start/end of the day to make sure that no courses get left out). Otherwise, the ranges most people would care about are too narrow.
- When the time filter is activated, it doesn't make the day/time filter actually show up as activated.

dlike230 (approved):
LGTM!

PR #315, opened 2021-10-31
Small course recs updates

dlike230:
- Apply bias to current courses (weight them more than past courses when computing recs)
- Determine whether to use prod model based on env variable


mureytasroc (approved):
LGTM

PR #317, opened 2021-11-05
Is open filter

AaDalal:
This pull request accomplishes the following:
- Adds `is_open_filter` to `CourseListSearch` viewset, which filters by courses for which all activities are open 
- Adds documentation for an additional query parameter`is_open` (with any argument) which adds the `is_open_filter`
- Tests `is_open_filter`


backend/courses/filters.py
```diff
@@ -176,6 +204,7 @@ def filter_queryset(self, request, queryset, view):
             "course_quality": bound_filter("course_quality"),
             "instructor_quality": bound_filter("instructor_quality"),
             "difficulty": bound_filter("difficulty"),
+            "is-open": is_open_filter,
```
mureytasroc, line 207: Might want to make this `is_open` for the sake of consistency

AaDalal: Good point

backend/tests/plan/test_api.py
```diff
@@ -116,6 +116,86 @@ def test_double_count_req(self):
...
+            status[0] for status in Section.STATUS_CHOICES if status[0] not in ["O"]
+        ]
+
+        self.client = APIClient()
+        set_semester()
+
+    def test_lec_open_all_rec_open(self):
+        response = self.client.get(reverse("courses-search", args=[TEST_SEMESTER]), {"is-open": ""})
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(len(response.data), 3)
```
mureytasroc, line 153: Shouldn't the response length always be 0 or 1 since there is only 1 course?

mureytasroc (approved):
Looks good! Just made some small comments and looks like you need to lint

backend/tests/plan/test_api.py
AaDalal: Yup my bad, updated in latest commit

PR #320, opened 2021-11-12
PCR filter plots by instructor

AaDalal:
- Adds `instructor_ids` query parameter to `course_plots` view, which accepts a comma separated string of instructor ids (ie the primary keys in the database). The course plots are filtered to only include data corresponding to one of the instructor ids. If the `instructor_ids` query parameter is an empty string, then the filter is not applied.
- Write tests to check if specifying instructors returns appropriate data
- Add appropriate autodocumentation for `course_plots`
- Added optional parameter `query_params` to `assertRequestContainsAppx`, which accepts a dictionary of key/value pairs representing query parameters and values (defaults to an empty dictionary)

**To do**
- [x] add additional tests for when a subset of instructors are selected
- [x] ~~tests for when invalid instructor ids are provided~~
- [x] Add autoschema docs


backend/review/views.py
```diff
@@ -186,23 +185,35 @@ def course_reviews(request, course_code):
...
+                        "in": "query",
+                        "description": "A comma-separated list of instructor IDs with which to filter the sections underlying the returned plots."  # noqa: E501
+                        "Note that if only invalid instructor IDs are present, plot response fields will be null or 0.",  # noqa: E501
+                        "schema": {"type": "string"},
+                        "required": False,
+                    },
+                ]
             },
         },
-        override_response_schema=course_plots_response_schema,
```
mureytasroc, line 205: It doesn't seem like you are changing the response schema here, any reason why this line should be changed?

backend/tests/review/test_stats.py
```diff
@@ -312,6 +312,14 @@ def test_course(self):
             "course-plots", "ESE-120", self.course_plots_subdict,
         )
 
+        instructor_ids = ",".join([str(id) for id in [Instructor.objects.get().pk]])
```
mureytasroc, line 315: Might also want to test for the case where only non-existent ID(s) are passed, to simultaneously test narrowing the data returned and the "no matches" edge case, using something like `empty_course_plots_subdict` from below

backend/tests/review/test_stats.py
```diff
@@ -739,6 +768,22 @@ def test_course(self):
             "course-plots", "ESE-120", self.course_plots_subdict,
         )
 
+        instructor_ids = ",".join(
```
mureytasroc, line 771: Could also use `",".join(Instructor.objects.values_list("id", flat=True))` here

mureytasroc (changes requested):
The only important thing that should be addressed here is the change from `course_plots_response_schema` to `course_reviews_response_schema` in `course_plots` (views.py). Otherwise, made some minor comments/suggestions, up to you whether you want to take them. Might be nice to add an extra test where you include some instructors but not others (like in `TwoInstructorsOneSectionTestCase`). Or you could just add this to the TODO at the bottom of the file lol.

mureytasroc (approved):

PR #323, opened 2021-11-14
Optimize meeting filter efficiency

mureytasroc:
This PR optimizes the day/time and is-open filters by using a mix of precomputation and better use of the Django ORM. Hopefully this leads to tractable performance in prod, otherwise we will likely have to simplify the queries to take any course for which some of its sections pass the relevant filters, as opposed to some section from each of its activities.


backend/courses/management/commands/load_status_history.py
```diff
@@ -94,7 +94,5 @@ def handle(self, *args, **kwargs):
 
             print(f"Finished loading status history from {src}... processed {row_count} rows. ")
 
-            print(
-                f"Recomputing PCA Demand Distribution Estimates for {len(semesters)} semesters..."
-            )
-            recompute_percent_open(semesters=",".join(semesters), verbose=True)
+            print(f"Recomputing PCA Stats for {len(semesters)} semesters...")
+            recompute_stats(semesters=",".join(semesters), verbose=True)
```
AaDalal, line 98: How often is this called? Is there a possible edge case where the fields added to the DB by `recompute_stats` and added in this PR are not actually present when the user first goes to add, for instance, a day time filter?

AaDalal: Actually looking at the migration, I think this is accounted for

AaDalal (approved):
Looks good! We discussed a couple things that we can implement if day time filter is still too slow (trying to do further optimization on the query and looking into some DB caching).

backend/courses/filters.py
```diff
@@ -23,30 +55,10 @@ def meeting_filter(queryset, meeting_query):
...
     """
-
-    matching_sections = Section.objects.filter(
-        id__in=Section.objects.annotate(num_meetings=Count("meetings"))
-        .filter(
-            num_meetings=subquery_count_distinct(
-                Meeting.objects.filter(meeting_query).filter(section_id=OuterRef("id")), column="id"
-            )
+    return queryset.filter(
+        id__in=course_ids_by_section_query(
```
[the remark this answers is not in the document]
mureytasroc: The issue is we also need a filter on the section `Q(num_meetings=0) | ...`. Also, I think this is the only place we filter course ids by meeting query? For the is_open filter, we filter course ids by section query right?

backend/courses/management/commands/load_status_history.py
mureytasroc: Yeah recompute_stats is run right after the registrar import (run in a cron job every ~24 hours), and I believe the registrar import is the only place that would add/remove sections or meetings.

backend/alert/management/commands/recomputestats.py
```diff
@@ -61,6 +65,45 @@ def get_semesters(semesters=None, verbose=False):
     return semesters
 
 
+def recompute_precomputed_fields(semesters=None, verbose=False, semesters_precomputed=False):
+    """
+    Recomputes Course.num_sections and Section.num_meetings fields for the given semesters.
```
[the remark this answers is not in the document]
mureytasroc: Nice catch

PR #329, opened 2021-11-21
Prevent registrating for open course

andyjiang3:


dlike230 (changes requested):
It would be more standard to do this with promises rather than a callback function. For example, you could do something like:
```
isCourseOpen.then(isOpen => {
    ...
})
```

Also, rather than using `sendError`, it would be better practice to call `setResponse` directly

frontend/alert/components/AlertForm/index.tsx
```diff
@@ -199,11 +224,15 @@ const AlertForm = ({
...
-                auto_resubscribe: autoResub === "true",
-            });
-            promises.push(promise);
+            isCourseOpen(section.section_id).then(isOpen => {
+                if (!isOpen) {
+                    const promise = doAPIRequest("/api/alert/registrations/", "POST", {
+                        section: section.section_id,
+                        auto_resubscribe: autoResub === "true",
+                    });
+                    promises.push(promise);
```
mureytasroc, line 233: I think we would want to put `const promise = isCourseOpen...` above and then `promises.push(promise);` below so that the `promises` array is definitely populated when we call `allSettled` below. Currently each promise is only added to the array once its `isCourseOpen` request returns, which causes a race condition between that and the interpreter reaching `Promise.allSettled`.

frontend/alert/components/AlertForm/index.tsx
```diff
@@ -180,17 +201,21 @@ const AlertForm = ({
             autoCompleteInputRef.current &&
             (autoCompleteInputRef.current.value === autofillSection || (autoCompleteInputRef.current.value !== "" && selectedCourses.size == 0))
         ) {
-            doAPIRequest("/api/alert/registrations/", "POST", {
-                section: autoCompleteInputRef.current.value,
-                auto_resubscribe: autoResub === "true",
+            isCourseOpen(autoCompleteInputRef.current.value).then(isOpen => {
+                if (!isOpen && autoCompleteInputRef.current) {
```
mureytasroc, line 205: This is minor, but maybe we could make a variable
```
const section = autoCompleteInputRef.current.value;
```
just below the if statement above, so we don't have to repeat this `autoCompleteInputRef.current` check. Also this would prevent a potential bug if the user changes the input field quickly after pressing submit (i.e. before the `isCourseOpen` request finishes).

mureytasroc (changes requested):
Nice work on this. There's just a minor bug that we should probably address before merging.

frontend/alert/components/AlertForm/index.tsx
```diff
@@ -136,7 +136,28 @@ const AlertForm = ({
         setResponse(new Response(blob, { status }));
     };
 
+    const isCourseOpen = (section) => {
+        return fetch(`/api/base/current/sections/${section}/`).then((res) => 
+            res.json().then((courseResult) => {
+                const blob = new Blob([JSON.stringify({message: "Course is currently open!", status: 400})], {
```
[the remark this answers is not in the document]
andyjiang3: Agree!

frontend/alert/components/AlertForm/index.tsx
andyjiang3: Oh thats right, good catch!

mureytasroc (approved):
Nice, LGTM!

Looks like there are missing semicolons in a couple places, but I think this is just an issue with our frontend auto-formatting / linting. I'll look into that in-depth at some point bc it seems to be a lot worse / less consistent than our setup for backend. Might be worth improving the setup and then making a big auto-formatting commit (sometime when we don't have many frontend PRs in the works).

PR #344, opened 2022-02-11
PCA Bulk Registration

andyjiang3:
Add bulk registration to PCA. 

- Registering for one section works very similar to previously
- Enter "bulk mode" when user selected more than one section. 
- In "bulk mode", user cannot edit input field & suggestions would be locked. 
- Click "X" on the side of input field or select only one section to exit out of "bulk mode"
- Added notifications when all selected sections are successfully registered

- Dynamic select all buttons - supports all types of activities (LEC, REC, IND, SEM, etc)

### DIFF
--- a/frontend/alert/components/AlertForm/index.tsx
+++ b/frontend/alert/components/AlertForm/index.tsx
@@ -5,9 +5,9 @@ import * as Sentry from "@sentry/browser";
 
 import { parsePhoneNumberFromString } from "libphonenumber-js/min";
 
+import { Center } from "pcx-shared-components/src/common/layout";
 import { Input } from "../Input";
 import AutoComplete from "../AutoComplete";
-import { Center } from "pcx-shared-components/src/common/layout";
 import getCsrf from "../../csrf";
 import { User } from "../../types";
 
--- a/frontend/alert/components/AlertForm/index.tsx
+++ b/frontend/alert/components/AlertForm/index.tsx
@@ -141,6 +141,11 @@ const AlertForm = ({
         );
     };
 
+    // Clear all sections the user selected
+    const clearSelections = () => {
+        setSelectedCourses(new Set());
+    }
+
     const submitRegistration = () => {
         //if user has a auto fill section and didn't change the input value then register for section
         if (autoCompleteInputRef.current && autoCompleteInputRef.current.value == autofillSection) {
@@ -164,9 +169,22 @@ const AlertForm = ({
             promises.push(promise)
         })
 
+        let success = true;
         Promise.all(promises)
-            .then((responses) => responses.map(res => setResponse(res)))
-            .catch(handleError);
+            .then((responses) => {
+                responses.forEach(res => setResponse(res));
+                success = false;
+                console.log(success);
+            })
+            .catch(handleError)
+            .finally(() => {
+                if (success) {
+                    sendError(201, "Successfully registered for all sections!")
+                    clearSelections();
+                }
+            })
+        
+        
     };
 
     const onSubmit = () => {
@@ -210,6 +228,7 @@ const AlertForm = ({
                 setSelectedCourses={setSelectedCourses}
                 setTimeline={setTimeline}
                 inputRef={autoCompleteInputRef}
+                clearSelections={clearSelections}
             />
             <Input
                 placeholder="Email"
--- a/frontend/alert/components/AlertForm/index.tsx
+++ b/frontend/alert/components/AlertForm/index.tsx
@@ -172,7 +172,7 @@ const AlertForm = ({
         let success = true;
         Promise.all(promises)
             .then((responses) => {
-                responses.forEach(res => setResponse(res));
+                responses.forEach((res: Response) => setResponse(res));
                 success = false;
                 console.log(success);
             })
--- a/frontend/alert/components/AlertForm/index.tsx
+++ b/frontend/alert/components/AlertForm/index.tsx
@@ -148,6 +148,13 @@ const AlertForm = ({
         setSelectedCourses(new Set());
     };
 
+    const deselectCourse = (section: Section): boolean => {
+        const newSelectedCourses = new Set(selectedCourses);
+        const removed = newSelectedCourses.delete(section);
+        removed && setSelectedCourses(newSelectedCourses);
+        return removed;
+    }
+
     const submitRegistration = () => {
         // if user has a auto fill section and didn't change the input value then register for section
         if (
@@ -172,20 +179,18 @@ const AlertForm = ({
             promises.push(promise);
         });
 
-        let success = true;
-        Promise.all(promises)
-            .then((responses) => {
-                responses.forEach((res: Response) => setResponse(res));
-                success = false;
-                console.log(success);
-            })
-            .catch(handleError)
-            .finally(() => {
-                if (success) {
-                    sendError(201, "Successfully registered for all sections!");
-                    clearSelections();
-                }
-            });
+        let sections = Array.from(selectedCourses)
+
+        Promise.allSettled(promises)
+            .then((responses) => responses.forEach(
+                (res: PromiseSettledResult<Response>, i) => {
+                    if (res.status === "fulfilled") {
+                        setResponse(res.value);
+                        deselectCourse(sections[i]);
+                    } else {
+                        handleError(res.reason);
+                    }
+                }));
     };
 
     const onSubmit = () => {
--- a/frontend/alert/components/AlertForm/index.tsx
+++ b/frontend/alert/components/AlertForm/index.tsx
@@ -159,7 +159,7 @@ const AlertForm = ({
         // if user has a auto fill section and didn't change the input value then register for section
         if (
             autoCompleteInputRef.current &&
-            autoCompleteInputRef.current.value == autofillSection
+            autoCompleteInputRef.current.value === autofillSection
         ) {
             doAPIRequest("/api/alert/registrations/", "POST", {
                 section: autofillSection,
@@ -179,12 +179,13 @@ const AlertForm = ({
             promises.push(promise);
         });
 
-        let sections = Array.from(selectedCourses)
+        const sections = Array.from(selectedCourses)
 
         Promise.allSettled(promises)
             .then((responses) => responses.forEach(
                 (res: PromiseSettledResult<Response>, i) => {
                     if (res.status === "fulfilled") {
+                        console.log("fulfilled")
                         setResponse(res.value);
                         deselectCourse(sections[i]);
                     } else {
--- a/frontend/alert/components/AlertForm/index.tsx
+++ b/frontend/alert/components/AlertForm/index.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from "react";
+import React, { useEffect, useState, useRef } from "react";
 import PropTypes from "prop-types";
 import styled from "styled-components";
 import * as Sentry from "@sentry/browser";
@@ -9,7 +9,7 @@ import { Center } from "pcx-shared-components/src/common/layout";
 import { Input } from "../Input";
 import AutoComplete from "../AutoComplete";
 import getCsrf from "../../csrf";
-import { User } from "../../types";
+import { User, Section } from "../../types";
 
 const SubmitButton = styled.button`
     border-radius: 5px;
@@ -101,13 +101,16 @@ const AlertForm = ({
     setTimeline,
     autofillSection = "",
 }: AlertFormProps) => {
-    const [section, setSection] = useState(autofillSection);
+    const [selectedCourses, setSelectedCourses] = useState<Set<Section>>(new Set())
+
     const [email, setEmail] = useState("");
 
     const [phone, setPhone] = useState("");
     const [isPhoneDirty, setPhoneDirty] = useState(false);
 
     const [autoResub, setAutoResub] = useState("false");
+
+    const autoCompleteInputRef = useRef<HTMLInputElement>(null);
 
     useEffect(() => {
         const phonenumber =
@@ -139,11 +142,30 @@ const AlertForm = ({
     };
 
     const submitRegistration = () => {
-        doAPIRequest("/api/alert/registrations/", "POST", {
-            section,
-            auto_resubscribe: autoResub === "true",
+        //if user has a auto fill section and didn't change the input value then register for section
+        if (autoCompleteInputRef.current && autoCompleteInputRef.current.value == autofillSection) {
+            doAPIRequest("/api/alert/registrations/", "POST", {
+                section: autofillSection,
+                auto_resubscribe: autoResub === "true",
+            })
+                .then((res) => setResponse(res))
+                .catch(handleError);
+        }
+
+        selectedCourses.forEach((section) => console.log(section.section_id))
+
+        // register all selected sections
+        let promises: any = []
+        selectedCourses.forEach((section) => {
+            const promise = doAPIRequest("/api/alert/registrations/", "POST", {
+                section: section.section_id,
+                auto_resubscribe: autoResub === "true",
+            })
+            promises.push(promise)
         })
-            .then((res) => setResponse(res))
+
+        Promise.all(promises)
+            .then((responses) => responses.map(res => setResponse(res)))
             .catch(handleError);
     };
 
@@ -184,8 +206,10 @@ const AlertForm = ({
         <Form>
             <AutoComplete
                 defaultValue={autofillSection}
-                onValueChange={setSection}
+                selectedCourses={selectedCourses}
+                setSelectedCourses={setSelectedCourses}
                 setTimeline={setTimeline}
+                inputRef={autoCompleteInputRef}
             />
             <Input
                 placeholder="Email"
--- a/frontend/alert/components/AlertForm/index.tsx
+++ b/frontend/alert/components/AlertForm/index.tsx
@@ -101,7 +101,9 @@ const AlertForm = ({
     setTimeline,
     autofillSection = "",
 }: AlertFormProps) => {
-    const [selectedCourses, setSelectedCourses] = useState<Set<Section>>(new Set())
+    const [selectedCourses, setSelectedCourses] = useState<Set<Section>>(
+        new Set()
+    );
 
     const [email, setEmail] = useState("");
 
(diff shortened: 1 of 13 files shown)


frontend/alert/components/AutoComplete.tsx
```diff
@@ -260,6 +213,13 @@ const AutoComplete = ({ defaultValue="", onValueChange, setTimeline, disabled })
         }
     }, [suggestionsFromBackend, value]);
 
+    // Create input text for bulk mode in the form of: [last selected section title] + [# selected courses - 1] more
+    const generateCoursesValue = () => {
+        let lastElement;
+        for (lastElement of Array.from(selectedCourses));
```
mureytasroc, line 219: Could just do `const lastElement = Array.from(selectedCourses).pop();`

frontend/alert/components/AutoComplete.tsx
```diff
@@ -238,12 +159,44 @@ const AutoComplete = ({ defaultValue="", onValueChange, setTimeline, disabled })
...
+                inputRef.current.value = generateCoursesValue();
+            } else if (selectedCourses.size == 1) {
+                // display the one selected section
+                setInternalValue(
+                    selectedCourses.values().next().value.section_id
+                );
+                inputRef.current.value = selectedCourses
+                    .values()
+                    .next().value.section_id;
+            } else if (!inputValDelete.current) {
```
mureytasroc, line 188: I'm probably just misunderstanding something, but why can't we put this logic down in the `onKeyUp` handler of `AutoCompleteInput` [here](https://github.com/pennlabs/penn-courses/pull/344/files#diff-196242b0d979a3169ad3a23cc53cc283728b32d9ce3ec855356930a209d9172cR328) and eliminate the `inputValDelete` ref (which is kind of confusing)?

frontend/alert/components/AutoComplete.tsx
```diff
@@ -238,12 +159,44 @@ const AutoComplete = ({ defaultValue="", onValueChange, setTimeline, disabled })
...
+                // No sections selected, clear input and suggestions
+                // Run when the only course that was selected is unselected
+                inputRef.current.value = "";
+                setInternalValue("");
+            } else {
+                inputValDelete.current = false;
+            }
+        }
+    }, [selectedCourses]);
```
mureytasroc, line 197: Noting that this is safe because we have the invariant that we only change `selectedCourses` by creating a new set, rather than mutating it (useEffect checks for referential changes in its dependencies).

frontend/alert/components/AutoComplete.tsx
```diff
@@ -283,6 +243,29 @@ const AutoComplete = ({ defaultValue="", onValueChange, setTimeline, disabled })
...
+        .reduce((res, obj) => {
+            const [courseName, midNum, endNum] = obj.section_id.split("-");
+            const { activity } = obj;
+
+            if (res[`${courseName}-${midNum}`]) {
+                if (res[`${courseName}-${midNum}`][activity]) {
+                    res[`${courseName}-${midNum}`][activity].push(obj);
+                } else {
+                    res[`${courseName}-${midNum}`][activity] = [obj];
+                }
```
mureytasroc, line 261: Could just be ``res[`${courseName}-${midNum}`][activity].push(obj);`` (eliminating the else case) since we know `res` values are always initialized to the form `{ [activity]: [obj] }`, right?

frontend/alert/components/GroupSuggestion.tsx
```diff
@@ -0,0 +1,357 @@
...
+            states[key] = false;
+        });
+
+        return states;
+    };
+
+    const [buttonStates, setButtonStates] = useState(initializeButtonStates);
+
+    /**
+     * Takes the activity/activity of a section and return if all sections of {activity} have been selected
```
mureytasroc, line 208: activity/activity?

frontend/alert/components/GroupSuggestion.tsx
```diff
@@ -0,0 +1,357 @@
...
+     */
+    const selectedAllActivity = (activity) => {
+        // check if the user selected all courses of a certain activity in a group suggestion
+        let selectedAll = true;
+
+        sections[activity].map((section) => {
+            if (!selectedCourses.has(section)) {
+                selectedAll = false;
+            }
+        });
```
mureytasroc, line 220: Could just do `const selectedAll = sections[activity].every((section) => selectedCourses.has(section));`

frontend/alert/components/GroupSuggestion.tsx
```diff
@@ -0,0 +1,357 @@
...
+    };
+
+    /**
+     * Update button states to match selected courses in the case the user is manually checking/unchecking sections
+     * Ex. allLectures should be true when user manually select all "LEC" sections
+     * @param activity - section activity
+     */
+    const syncButtonStates = (activity, correctState) => {
+        // update button state if its not the same as correctState
+        if (buttonStates[activity] != correctState) {
```
mureytasroc, line 233: I think `!==` is considered better practice

frontend/alert/components/GroupSuggestion.tsx
```diff
@@ -0,0 +1,357 @@
...
+     * Takes the activity of button clicked: LEC, REC, LABS, etc and switch state
+     * @param activity - section activity
+     */
+    const toggleButton = (activity) => {
+        const newButtonStates = { ...buttonStates };
+
+        newButtonStates[activity] = !newButtonStates[activity];
+        syncCheckAlls(activity, newButtonStates[activity], setSelectedCourses);
+
+        setButtonStates(newButtonStates);
```
mureytasroc, line 250: Is there any issue with
```typescript
const newState = !buttonStates[activity];
syncButtonStates(activity, newState);
syncCheckAlls(activity, newState, setSelectedCourses);
```

frontend/alert/components/GroupSuggestion.tsx
```diff
@@ -0,0 +1,357 @@
...
+    };
+
+    /**
+     * If one of the select all is toggled, update selected courses to match
+     * @param activity - section activity
+     * @param checkedAll - whether all sections of activity should be checked or not
+     * @param setSelectedCourses - method to update selected courses set
```
mureytasroc, line 257: Any reason why this function needs to take `setSelectedCourses` as an argument, but `toggleButton` doesn't?

frontend/alert/components/GroupSuggestion.tsx
```diff
@@ -0,0 +1,357 @@
...
+            });
+            setSelectedCourses(newSelectedCourses);
+        }
+    };
+
+    /**
+     * Update the selected courses set when user check/uncheck box
+     * @param selectedCourses - selected courses set
+     * @param setSelectedCourses - method to update selected courses set
+     * @param setSelectedCourses - the section
```
mureytasroc, line 281: I think this should be `suggestion` not `setSelectedCourses` (duplicated param)?

frontend/alert/components/GroupSuggestion.tsx
```diff
@@ -0,0 +1,357 @@
...
+    };
+
+    /**
+     * If one of the select all is toggled, update selected courses to match
+     * @param activity - section activity
+     * @param checkedAll - whether all sections of activity should be checked or not
+     * @param setSelectedCourses - method to update selected courses set
+     */
+    const syncCheckAlls = (activity, checkedAll, setSelectedCourses) => {
+        const newSelectedCourses = new Set(selectedCourses);
```
mureytasroc, line 260: Could this go inside the if statement below?

frontend/alert/components/GroupSuggestion.tsx
```diff
@@ -0,0 +1,357 @@
...
+            sections[activity].map((section) => {
+                // checkedAll = false but selectedCourses has course
+                if (newSelectedCourses.has(section) && !checkedAll) {
+                    newSelectedCourses.delete(section);
+
+                    // checkedAll = true but selectedCourses doesn't have course
+                } else if (!newSelectedCourses.has(section) && checkedAll) {
+                    newSelectedCourses.add(section);
+                }
+            });
```
mureytasroc, line 272: Could we replace the above with
```typescript
sections[activity].forEach(checkedAll ? newSelectedCourses.add : newSelectedCourses.delete);
```

frontend/alert/components/GroupSuggestion.tsx
```diff
@@ -0,0 +1,357 @@
...
+            setSelectedCourses(newSelectedCourses);
+        }
+    };
+
+    /**
+     * Update the selected courses set when user check/uncheck box
+     * @param selectedCourses - selected courses set
+     * @param setSelectedCourses - method to update selected courses set
```
mureytasroc, line 280: Same question about the `setSelectedCourses` parameter as above

mureytasroc (changes requested):
Really awesome work with this PR! This is a large feature and I think it will really improve PCA. Just made some suggestions / left a question.

frontend/alert/components/AlertForm/index.tsx
```diff
@@ -101,14 +101,19 @@ const AlertForm = ({
     setTimeline,
     autofillSection = "",
 }: AlertFormProps) => {
-    const [section, setSection] = useState(autofillSection);
+    const [selectedCourses, setSelectedCourses] = useState<Set<Section>>(
+        new Set()
```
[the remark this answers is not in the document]
andyjiang3: Yeah I thought of that but `Set` is type `Section` while` autofillSection` is `string` so I thought it would be easier to just add a special case. Is there anything we can do instead that would work be better?

frontend/alert/components/AlertForm/index.tsx
```diff
@@ -138,13 +143,49 @@ const AlertForm = ({
...
+        // if user has a auto fill section and didn't change the input value then register for section
+        if (
+            autoCompleteInputRef.current &&
+            autoCompleteInputRef.current.value == autofillSection
+        ) {
+            doAPIRequest("/api/alert/registrations/", "POST", {
+                section: autofillSection,
+                auto_resubscribe: autoResub === "true",
+            })
+                .then((res) => setResponse(res))
```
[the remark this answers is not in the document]
andyjiang3: Yup!

frontend/alert/components/GroupSuggestion.tsx
andyjiang3: Originally named course activity as course type but then decided it would be confusing so I replaced all. This is probably a mistake from replacing all type to activity.

frontend/alert/components/GroupSuggestion.tsx
andyjiang3: Ah thats very true!

frontend/alert/components/GroupSuggestion.tsx
andyjiang3: Right!!

frontend/alert/components/GroupSuggestion.tsx
andyjiang3: Yeah, I actually don't think the if statement is necessary since `sections[activity]` will always be defined in this case because of the way `groupSuggestions` is created and `syncCheckAlls` is only called on existing activities within the `sections`.

frontend/alert/components/GroupSuggestion.tsx
andyjiang3: Oops, I didn't need to pass `setSelectedCourses` into `syncCheckAlls`.

frontend/alert/components/AlertForm/index.tsx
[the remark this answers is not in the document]
mureytasroc: Hmm yeah ig that's fine then. Maybe it would be worth doing an API request for the info of the default section, but that would prob just add extra lag. The special case logic isn't too substantial anyway.

frontend/alert/components/GroupSuggestion.tsx
andyjiang3: Yeah for sure! Definitely didn't need the `has` checks

frontend/alert/components/AlertForm/index.tsx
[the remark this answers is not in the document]
andyjiang3: Sounds good!

frontend/alert/components/AutoComplete.tsx
andyjiang3: Doing this gives me `Type 'Section | undefined' is not assignable to type 'Section', Type 'undefined' is not assignable to type 'Section'` (although lastElement cannot be undefined..) which is why I decided to loop over it. Any suggestions on how to fix this error?

frontend/alert/components/AutoComplete.tsx
andyjiang3: This portion of the code should only be run when the user uncheck the only course they have selected (which causes the value of the input to empty). However, since I am calling `clearSelection` (which modify `selectedCourses`) in `onKeyUp`, this portion of the code will be triggered when it is not suppose to (which causes pressing the backspace to remove the entire value of the input). So I needed to make sure the user is not pressing backspace. Lmk if this make sense lol

Totally agree inputValDelete is confusing, definitely open to any suggestions on how to make this more readable.

frontend/alert/components/AutoComplete.tsx
mureytasroc: Hmm then maybe `const lastElement = Array.from(selectedCourses).pop()!;` would work? This is the [non-null assertion operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator) (works for `undefined` too).

frontend/alert/components/AutoComplete.tsx
andyjiang3: Hmm, I'm basically grouping all the sections by its course and then grouping by its activity (subgroups) within each group. Something like this:

`CIS-120:
     LEC: [{CIS-120-100, CIS-120-101}]
     REC: [{CIS-120-110, CIS-120-111}]
`

`res[`${courseName}-${midNum}`] = { [activity]: [obj] }` would create a new group (for a new course) and  `res[`${courseName}-${midNum}`][activity] = [obj]` would create a new subgroup (for a new activity that current isnt a subgroup) within each group. Just having `res[`${courseName}-${midNum}`][activity].push(obj)` wouldn't work when its a new activity that isn't a subgroup yet right? I could totally be wrong lol

Is there a better way of grouping the sections into this format?

mureytasroc: Oh my bad, you're totally right.

frontend/alert/components/AutoComplete.tsx
mureytasroc: This makes sense... but maybe passing a `clearInput` function to the `GroupSuggestion` component instead of clearing the input here would allow us to eliminate this confusing code? It makes sense once explained, I'm just worried that someone in the future reading this will have a hard time figuring out what's going on.

frontend/alert/components/AutoComplete.tsx
andyjiang3: ahh didn't know that, thanks!

frontend/alert/components/AlertForm/index.tsx
```diff
@@ -138,13 +143,49 @@ const AlertForm = ({
...
+                success = false;
+                console.log(success);
+            })
+            .catch(handleError)
+            .finally(() => {
+                if (success) {
+                    sendError(201, "Successfully registered for all sections!");
+                    clearSelections();
+                }
+            });
```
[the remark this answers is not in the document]
andyjiang3: Great suggestion! One thing, is it suppose to send a notification and remove the course from `selectedCourses` when a section has been successfully registered? For some reason it'll only show notification if theres an error. I had this problem before and it seemed like registrations that are successful didn't send any responses, which is really weird. I tested on the master branch as well and it didn't send a notification if a section is successfully registered. Do you know why this is happening? I pushed a commit with the changes above.

frontend/alert/components/AutoComplete.tsx
andyjiang3: Oh the thing is I don't want to clear the input when the user is pressing the backspace. I just want to remove the selected course from `selectedCourses`. But doing so triggers the `useEffect` which would clear the input value.

The input value should only clear if the user deselect the only course they have selected (which is why its handled inside the `useEffect`). So I need the `inputValDelete` in the `useEffect` to separate whether it is triggered because the user pressed backspace vs the user deselect the only course they have selected (since both actions modify `selectedCourses` and thus trigger the `useEffect`)

mureytasroc: But if we passed a `clearInput` function to `GroupSuggestion`, couldn't we handle the case where the user deselects the only course they have selected inside `GroupSuggestion`, and then get rid of these extra cases (for `selectedCourses.size === 0` in the `useEffect`)? That way we wouldn't have to check for `inputValDelete` because these cases wouldn't be handled in the `useEffect`?

frontend/alert/components/AlertForm/index.tsx
```diff
@@ -138,13 +143,54 @@ const AlertForm = ({
...
+        const promises: Array<Promise<Response>> = [];
+        selectedCourses.forEach((section) => {
+            const promise = doAPIRequest("/api/alert/registrations/", "POST", {
+                section: section.section_id,
+                auto_resubscribe: autoResub === "true",
+            });
+            promises.push(promise);
+        });
+
+        let sections = Array.from(selectedCourses)
```
mureytasroc, line 182: Could be `const`

frontend/alert/components/AutoComplete.tsx
```diff
@@ -238,12 +168,40 @@ const AutoComplete = ({ defaultValue="", onValueChange, setTimeline, disabled })
...
+            initialRender.current = false;
+            return;
+        }
 
+        // update suggestions and input value
+        if (inputRef.current) {
+            if (bulkMode) {
+                // display all sections selected
+                inputRef.current.value = generateCoursesValue();
+            } else if (selectedCourses.size == 1) {
```
mureytasroc, line 185: Should be `===`

frontend/alert/components/AutoComplete.tsx
```diff
@@ -292,105 +281,103 @@ const AutoComplete = ({ defaultValue="", onValueChange, setTimeline, disabled })
...
+                        } else if (e.keyCode === UP_ARROW && suggestions) {
+                            // select a suggestion when the up arrow key is pressed
+                            const newSelectedSuggestion = Math.max(
+                                getCurrentIndex() - 1,
+                                -1
                             );
+                            handleSuggestionSelect(newSelectedSuggestion);
+                        } else if (
+                            e.keyCode === DELETE_KEY &&
+                            selectedCourses.size == 1
```
mureytasroc, line 331: Should be `===`

frontend/alert/components/AlertForm/index.tsx
```diff
@@ -138,13 +143,54 @@ const AlertForm = ({
...
-        doAPIRequest("/api/alert/registrations/", "POST", {
-            section,
-            auto_resubscribe: autoResub === "true",
-        })
-            .then((res) => setResponse(res))
-            .catch(handleError);
+        // if user has a auto fill section and didn't change the input value then register for section
+        if (
+            autoCompleteInputRef.current &&
+            autoCompleteInputRef.current.value == autofillSection
```
mureytasroc, line 162: Should be `===`

mureytasroc (approved):
There are still a few unresolved comments, but this should be good to merge at your discretion. Might want to wait until Dan has a chance to take a look (but prob fine to merge without his review if he's too busy).

mureytasroc:
🎉🎉🎉

PR #349, opened 2022-03-12
Course Topics

mureytasroc:
This PR adds the `Topic` model to the PCx backend, which organizes related `Course` objects across different semesters (see [this Notion page](https://www.notion.so/pennlabs/Tracking-Courses-Over-Time-72dc45bb801349e9af10e6123e6cb690) for more about the motivation and design of this feature). Previously this linking was done implicitly by course code. However, to accommodate all the course code changes taking place in spring 2022 (as part of the NGSS transition), we added the `Topic` model as a more general abstraction.

This PR also changes our PCR routes to group courses by topic, rather than code (without changing expected request schemas). It also adds some useful fields to the response schemas of PCR routes (explained in the API documentation: `localhost:8000/api/documentation`). These fields will allow the frontend to show historical codes representing a topic and (optionally) a `branched_from` code in the case of topic branches.

It also adds code to the `Course.save` method and a few manually-runnable scripts to automatically link courses into topics. The scripts can take user input on whether certain courses (detected as similar but not passing thresholds for automatic linking) should be joined into the same topic. Finally, these scripts can be passed a crosswalk of course code changes (as provided by [ISC](https://bit.ly/3HtqPq3)).


backend/courses/management/commands/link_courses_to_topics.py
```diff
@@ -0,0 +1,215 @@
...
+    """
+    From a given crosswalk csv path, generate a dict mapping new_full_code->old_full_code,
+    ignoring branched links in the crosswalk (a course splitting into multiple new courses).
+    """
+    guaranteed_links = dict()
+    cross_walk = pd.read_csv(cross_walk, delimiter="|", encoding="unicode_escape")
+    for _, r in cross_walk.iterrows():
+        old_full_code = f"{r['SRS_SUBJ_CODE']}-{r['SRS_COURSE_NUMBER']}"
+        new_full_code = f"{r['NGSS_SUBJECT']}-{r['NGSS_COURSE_NUMBER']}"
+        if old_full_code in guaranteed_links:
```
AaDalal, line 47: Aren't `old_full_code`s only in the value set of `guaranteed_links`?

backend/courses/management/commands/link_courses_to_topics.py
```diff
@@ -0,0 +1,215 @@
...
+            # Ignore branched links
+            del guaranteed_links[new_full_code]
```
AaDalal, line 49: Could this lead to oscillatory behavior where if you have an even number of branches then there are no associations in the dict, but if you have an odd number of links there is an association?

PR #350, opened 2022-03-16
Review averages across semesters and departments

AaDalal:
This pull request adds a command to aggregate averages by semester and department.

### DIFF
--- a/backend/courses/management/commands/compute_review_semester_department.py
+++ b/backend/courses/management/commands/compute_review_semester_department.py
@@ -29,7 +29,7 @@ def average_by_dept(fields, semesters="all", departments=None, path=None):
             fields=fields,
             subfilters={
                 "review__section__course__semester": semester,
-                "review__section__course__department__id": OuterRef("id"),
+                "review__section__course__department_id": OuterRef("id"),
             },
             extra_metrics=False,
         ).values("code", *fields)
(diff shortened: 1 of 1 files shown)


backend/courses/management/commands/compute_review_semester_department.py
```diff
@@ -0,0 +1,77 @@
...
+    Note that if fields should be a list of strings representing the review fields to be aggregated.
+    """
+    dept_avgs = {}
+    for semester in tqdm(get_semesters(semesters="all")):
+        semester_dept_avgs = review_averages(
+            Department.objects.all(),
+            fields=fields,
+            subfilters={
+                "review__section__course__semester": semester,
+                "review__section__course__department__id": OuterRef("id"),
```
mureytasroc, line 27: Small detail but you can access the `id` field with `department_id` rather than `department__id` to avoid a join.

backend/courses/management/commands/compute_review_semester_department.py
```diff
@@ -0,0 +1,77 @@
...
+            type=str,
+            help=dedent(
+                """
+                path to the output file. If not provided then will simply be printed to console.
+                """
+            ),
+        )
+
+    def handle(self, *args, **kwargs):
+        fields = re.split(r"\w", kwargs["fields"])
```
mureytasroc, line 74: Might be better to have the fields be comma-separated and then you can just do `kwargs["fields"].strip().split(",")`. Any reason why you opted to split by whitespace here?

mureytasroc (approved):

backend/courses/management/commands/compute_review_semester_department.py
AaDalal: Good point. I'll update that.

backend/courses/management/commands/compute_review_semester_department.py
AaDalal: Interesting. I'll make the change

PR #351, opened 2022-03-19
Bulk Reg - Support Manual Reg + Bug Fix

andyjiang3:
• Allow user to reg for section manually typed into input field (without checking checkbox)
• Clear input field if all selected courses are successfully registered
• Fix bug where unsuccessfully registered courses was been deselected from `selectedCourses` set

### DIFF
--- a/frontend/alert/components/AlertForm/index.tsx
+++ b/frontend/alert/components/AlertForm/index.tsx
@@ -104,6 +104,7 @@ const AlertForm = ({
     const [selectedCourses, setSelectedCourses] = useState<Set<Section>>(
         new Set()
     );
+    const [value, setValue] = useState(autofillSection);
 
     const [email, setEmail] = useState("");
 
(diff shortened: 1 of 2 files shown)


frontend/alert/components/AlertForm/index.tsx
```diff
@@ -184,10 +211,15 @@ const AlertForm = ({
         Promise.allSettled(promises)
             .then((responses) => responses.forEach(
                 (res: PromiseSettledResult<Response>, i) => {
+                    //fulfilled if response is returned, even if reg is unsuccessful.
                     if (res.status === "fulfilled") {
-                        console.log("fulfilled")
-                        setResponse(res.value);
-                        deselectCourse(sections[i]);
+                        if (res.value.status === 200) {
```
mureytasroc, line 216: Could just be:
```
setResponse(res.value);
if (res.value.ok) {
    deselectCourse(sections[i]);
}
```

frontend/alert/components/AlertForm/index.tsx
```diff
@@ -184,10 +211,15 @@ const AlertForm = ({
         Promise.allSettled(promises)
             .then((responses) => responses.forEach(
                 (res: PromiseSettledResult<Response>, i) => {
+                    //fulfilled if response is returned, even if reg is unsuccessful.
```
mureytasroc, line 214: Turns out this behavior (accept on any response code) comes from `fetch` (used in `doAPIRequest`): https://stackoverflow.com/a/38236296
This makes sense, because looking at the code above `handleError` displays an error message saying:
> "We're sorry, but there was an error sending your request to our servers. Please try again!"

so it wouldn't make sense for `handleError` to be called in the case of non-2xx responses.

frontend/alert/components/AlertForm/index.tsx
```diff
@@ -184,10 +211,15 @@ const AlertForm = ({
...
-                        console.log("fulfilled")
-                        setResponse(res.value);
-                        deselectCourse(sections[i]);
+                        if (res.value.status === 200) {
+                            setResponse(res.value);
+                            deselectCourse(sections[i]);
+                        } else {
+                            setResponse(res.value);
+                        }
+                    //only if error occurred
```
mureytasroc, line 222: could say "only if network error occured" (see stack overflow post above) to disambiguate with non-2xx responses, which might be confused with an "error"

frontend/alert/components/AlertForm/index.tsx
```diff
@@ -148,25 +149,51 @@ const AlertForm = ({
...
             doAPIRequest("/api/alert/registrations/", "POST", {
-                section: autofillSection,
+                section: autoCompleteInputRef.current.value,
                 auto_resubscribe: autoResub === "true",
             })
-                .then(setResponse)
+                .then((res) => {
+                    if (res.status === 200) {
+                        clearInputValue();
+                    }
```
mureytasroc, line 190: See below, this should probably be:
```
if (res.ok) { 
    clearInputValue();
}
```
The [Create Registration](https://penncoursealert.com/api/documentation/#operation/Create%20Registration) route returns 201 on success (not 200). The `ok` field of a `Response` object is true iff the response code is 2xx.

mureytasroc (approved):
Looks good! Just a couple issues which should be resolved before merging (made comments below), but marking this as approved so you can merge at your discretion.

frontend/alert/components/AlertForm/index.tsx
andyjiang3: Ah that make sense!

andyjiang3:
@mureytasroc Looks like I dont have permission to merge PRs but I resolved the issues!

PR #358, opened 2022-04-02
Use Kittyhawk for Deployment Config

joyliu-q:
This PR does the following:
- Transition from icarus to kittyhawk
- Update kraken and regenerate deployment CI
- Switch backend from `DOMAIN` to `DOMAINS` (list of domains supported)

To generate yaml, go into the `k8s` folder and run:
- `yarn build`

Checklist
- [x] package.json in `k8s` folder is uniform
- [x] package.json in `.github/cdk` folder is uniform 
- [x] renamed any mentions to `DOMAIN` in backend to `DOMAINS`
- [x] Update kittyhawk version after ingress port bug PR gets merged
- [x] Run `yarn build` in `k8s` folder and compare new generated manifests w current


.github/cdk/package.json
```diff
@@ -15,7 +15,7 @@
   "dependencies": {
     "@pennlabs/kraken": "^0.8.6",
     "cdkactions": "^0.2.3",
-    "constructs": "^3.3.263"
+    "constructs": "^3.2.109"
```
joyliu-q, line 18: This is the version used by `cdkactions`. Using a `3.3.x` version would cause some products (e.g. penn-clubs to break)

joyliu-q:
```
main.ts:6:26 - error TS2345: Argument of type 'App' is not assignable to parameter of type 'Construct'.
  Property 'onValidate' is protected but type 'Construct' is not a class derived from 'Construct'.
  ```

PR #365, opened 2022-04-05
Temporary: revert Topic changes to queries

mureytasroc:
Not a permanent fix but this will hopefully get all our products working again (will continue to try to reproduce / debug locally).


mureytasroc:
Not a permanent fix but this will hopefully get all our products working again (will continue to try to reproduce / debug locally).

PR #378, opened 2022-04-13
Better instructions to debug docker compose up

AaDalal:
Add debugging steps for `docker compose`.


mureytasroc:
I think the more recent change we made covers the instructions to start docker manually, and it's probably unnecessary to put more detailed debugging instructions in the README (we can help future members directly if they run into issues like these). I could be missing something though -- lmk if so.

PR #385, opened 2022-05-11
Change PCA Alert Text to Path@Penn

andyjiang3:
Change text from PennInTouch to Path@Penn. Used https://twiliodeved.github.io/message-segment-calculator/ to make sure text is 1 segment. 

Also should we consider deleting "- Thanks for using Penn Course Alert!" since sometimes the message can be 2 segments because of new 4-digit course code.


mureytasroc (approved):
Looks good, nice catch -- I don't think it's necessary to remove "Thanks for using Penn Course Alert" because in all likelihood we will soon be removing the resubscribe text anyway (if we are moving to just always using auto-resubscribe), which would offset increased text length from the longer course code.

PR #402, opened 2022-08-28
Pcp schedule sharing

anshnagwekar:
Outlined the API for the schedule sharing features (added a Friendship model and is_shared field in Schedule model


backend/courses/models.py
```diff
@@ -1419,3 +1419,40 @@ def create_or_update_user_profile(sender, instance, created, **kwargs):
...
     instance.profile.save()
+
+
+
+class Friendship(models.Model):
+    """
+    Used to track friendships along with requests status 
+    """
+    
+    requestSender = models.ForeignKey(
```
mureytasroc, line 1430: snake_case is typically considered to be more idiomatic for Python (according to PEP 8): https://peps.python.org/pep-0008/#function-and-variable-names

backend/courses/models.py
```diff
@@ -1419,3 +1419,40 @@ def create_or_update_user_profile(sender, instance, created, **kwargs):
...
+    )
+
+    requestReciever = models.ForeignKey(
+        UserProfile,
+        related_name="receiver",
+         on_delete=models.CASCADE,
+        help_text="The person (user) who recieved the request.",
+    )
+
+    status = models.IntegerField(
```
mureytasroc, line 1444: Could we replace this with an enum of `SENT`, `ACCEPTED`, and `REJECTED` ? Django has built-in support for enums: https://stackoverflow.com/a/58051918

backend/courses/models.py
```diff
@@ -1419,3 +1419,40 @@ def create_or_update_user_profile(sender, instance, created, **kwargs):
...
+        help_text=dedent(
+            """
+        Whether the request was sent or accepted (0 if sent, 
+        1 if recieved and accepted, deleted if rejected)
+        """
+        ),
+    )
+
+    accepted_at = models.DateTimeField(auto_now_add=True)
```
mureytasroc, line 1453: I don't think we can just use `auto_now_add` here if we make the enum change suggested in the comment above, unfortunately. Might need to override the save method: https://ilovedjango.com/django/models-and-databases/override-save-method-django-models/

backend/courses/models.py
```diff
@@ -1419,3 +1419,40 @@ def create_or_update_user_profile(sender, instance, created, **kwargs):
...
+    sent_at = models.DateTimeField()
+
```
mureytasroc, line 1455: As mentioned in a comment above, we might want to define a unique constraint here:
```
class Meta:
        unique_together = (("request_sender", "request_recipient"),)
```

backend/courses/models.py
```diff
@@ -1419,3 +1419,40 @@ def create_or_update_user_profile(sender, instance, created, **kwargs):
...
+    """
+    
+    requestSender = models.ForeignKey(
+        UserProfile,
+        on_delete=models.CASCADE,
+        related_name="sender",
+        help_text="The person (user) who sent the request.",
+    )
+
+    requestReciever = models.ForeignKey(
```
mureytasroc, line 1437: This is _super_ pedantic, but we might want to call this `request_recipient` (as a receiver is usually a device that handles receiving, as opposed to a person receiving a single request).

backend/courses/models.py
```diff
@@ -1419,3 +1419,40 @@ def create_or_update_user_profile(sender, instance, created, **kwargs):
...
+        1 if recieved and accepted, deleted if rejected)
+        """
+        ),
+    )
+
+    accepted_at = models.DateTimeField(auto_now_add=True)
+    sent_at = models.DateTimeField()
+
+    def __str__(self):
+        return "Sender: %s, Reciever: %s, Status: %s" % (self.requestSender, self.requestReciever, self.status)
```
mureytasroc, line 1457: An f string might be a big cleaner here: https://www.digitalocean.com/community/tutorials/python-f-strings-literal-string-interpolation

backend/courses/views.py
```diff
@@ -376,3 +378,43 @@ def get_queryset(self):
             section__course__semester=get_current_semester(),
             in_add_drop_period=True,
         ).order_by("created_at")
+
+
+# Function-Based View for Friendship Handling
+def sendFriendshipRequest(request):
+    res = {}
+    # verify that the friendship request can be sent (only if it doesn't already exist and
```
mureytasroc, line 386: Nice defensive mindset here - if we add the unique constraint above, we can just use `friendship, created = Friendship.get_or_create(...)` and return an error if `created` is false: 
https://docs.djangoproject.com/en/4.1/ref/models/querysets/#get-or-create

backend/courses/views.py
```diff
@@ -376,3 +378,43 @@ def get_queryset(self):
...
+    # if the requestor and sender exist)
```
mureytasroc, line 387:
```
from django.shortcuts import get_object_or_404
sender = get_object_or_404(User, id=request.user.id)
recipient = get_object_or_404(User, id=recipient_id)
```

backend/courses/views.py
```diff
@@ -376,3 +378,43 @@ def get_queryset(self):
...
+
```
mureytasroc, line 388: We might also want to check if a request exists in the reverse direction, i.e.
```python
from rest_framework import status
from django.forms.models import model_to_dict

try:
    existing_request = Friendship.get(
        recipient=request.user.id, sender=recipient_id)
    existing_request.status = Friendship.Status.ACCEPTED
    existing_request.save()
    return JsonResponse(
        {"message": "Friend request was in reverse direction already existed - friendship confirmed."},
        status = status.HTTP_202_ACCEPTED)
except Friendship.DoesNotExist:
    pass
...
return JsonResponse(model_to_dict(friendship), status.HTTP_202_ACCEPTED)
```

mureytasroc (commented):
Nice job so far!

PR #403, opened 2022-09-04
Patch PCP mobile

AaDalal:
- Restyles tabs
- Rescales UI to fit mobile
  - Increase search bar size
  - increase font size 
- Makes other small visual changes

This PR is a stopgap to help make PCP mobile more usable.

TODO:
- [x] add back PCP logo
- [x] ~~change location of filters~~
- [x] force switch to search tab when search bar is clicked
- [x] Test schedule tab (with courses that have times associated)
- [x] ~~Fix weird multiple font issue~~

### DIFF
--- a/frontend/plan/components/schedule/Schedule.tsx
+++ b/frontend/plan/components/schedule/Schedule.tsx
@@ -86,6 +86,9 @@ const ScheduleBox = styled.div`
     display: block;
     padding: 1.25rem;
     min-height: calc(100vh - 12em);
+    @media only screen and (min-width: 769px) {
+        height: calc(100vh - 12em);
+    }
 `;
 
 const ScheduleContents = styled.div`
--- a/frontend/plan/components/schedule/Schedule.tsx
+++ b/frontend/plan/components/schedule/Schedule.tsx
@@ -85,7 +85,7 @@ const ScheduleBox = styled.div`
     color: #4a4a4a;
     display: block;
     padding: 1.25rem;
-    height: calc(100vh - 12em);
+    min-height: calc(100vh - 12em);
 `;
 
 const ScheduleContents = styled.div`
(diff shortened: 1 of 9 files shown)


frontend/plan/components/search/SearchBar.tsx
```diff
@@ -5,6 +5,8 @@ import styled from "styled-components";
 // @ts-ignore
 import AccountIndicator from "pcx-shared-components/src/accounts/AccountIndicator";
 import { useRouter } from "next/router";
+import * as ReactDOM from "react-dom";
+import dynamic from "next/dynamic";
```
andyjiang3, line 9: Are these unused imports? If so, can you remove them

frontend/plan/components/schedule/Schedule.tsx
```diff
@@ -85,7 +85,7 @@ const ScheduleBox = styled.div`
     color: #4a4a4a;
     display: block;
     padding: 1.25rem;
-    height: calc(100vh - 12em);
+    min-height: calc(100vh - 12em);
 `;
```
andyjiang3, line 89: Seems like adding a course to the schedule expands the schedule box height, which makes the height of the boxes unequal. Why do we need `min-height` here?

Screen Shot 2022-09-06 at 8 32 53 PM

andyjiang3 (commented):
Great job resolving the mobile issues! Overall it looks good, just some comments! 

Also, the fonts and icons look kinda small on mobile view (might be because im looking at it on a laptop), maybe thats something we can revisit later cuz seems like we need to scale everything up a bit on mobile. 

Screen Shot 2022-09-06 at 8 37 19 PM

frontend/plan/components/search/SearchBar.tsx
AaDalal: Yup, good catch.

frontend/plan/components/schedule/Schedule.tsx
AaDalal: min height allows for expansion on mobile, where the viewport height may not be sufficient to hold the stats at the bottom. I'll create a new styled component for mobile or add a media query.

andyjiang3 (approved):
LGTM!

PR #405, opened 2022-09-13
Pcp schedule sharing (with flushed out views and urls)

anshnagwekar:
See comments in code for any clarification!


backend/plan/urls.py
```diff
@@ -12,5 +17,7 @@
 urlpatterns = [
     path("", TemplateView.as_view(template_name="plan/build/index.html")),
     path("recommendations/", recommend_courses_view, name="recommend-courses"),
+    path("shared/get_schedule", get_shared_schedule, name="get-shared-schedule"),
```
mureytasroc, line 20: A more RESTful route would be GET "schedules/shared" (and we could accept "user_id" as a query parameter to optionally filter shared schedules by a specific friend user -- otherwise we can return all schedules shared with the current user).

Including "get" in the route name is kind of redundant because the method name is GET.

backend/plan/urls.py
```diff
@@ -12,5 +17,7 @@
...
+    path("shared/set_schedule", set_shared_schedule, name="set-shared-schedule"),
```
mureytasroc, line 21: Do we need a route for this? Couldn't we just use the existing update schedule route to directly set `is_shared = True`? If so, we might want to add some validation to [that route](https://github.com/pennlabs/penn-courses/blob/494cb5326617a02363afa9267b064de7f6d16612/backend/plan/views.py#L385) (create + update schedule) to return a better error message if the at most one shared schedule constraint is violated (rather than bubbling up a 500 SQL exception), and document that case causing a 400 [here](https://github.com/pennlabs/penn-courses/blob/494cb5326617a02363afa9267b064de7f6d16612/backend/plan/views.py#L267), just like the semester check. Don't think we would need to even change the [serializer](https://github.com/pennlabs/penn-courses/blob/494cb5326617a02363afa9267b064de7f6d16612/backend/plan/serializers.py), as all fields are included by default.

mureytasroc (commented):
Nice work - feel free to ignore anything if it's too pedantic (currently experiencing the annoyance of a too-pedantic reviewer coworker at work and it can be counterproductive imo -- but also I learn a lot from their comments so I think at least seeing the comments is a benefit).

backend/plan/urls.py
anshnagwekar: Should we keep the get-shared-schedule route to ensure that we can get the shared schedule? I do agree that we can handle the set-shared-schedule tasks in the update route for the class view of Schedule.

backend/plan/urls.py
anshnagwekar: I like this design a lot! Will be implementing this, thanks for the suggestion!

backend/plan/urls.py
mureytasroc: Yeah I think we do need a separate route for getting shared schedules, bc the GET method on the schedules viewset will return the user's schedules, not schedules that are shared with the user.

backend/courses/views.py
```diff
@@ -376,3 +384,122 @@ def get_queryset(self):
...
+
+# private helper function used to get user's accepted friends
+def get_accepted_friends(user):
+    friendships = Friendship.objects.filter(Q(sender=user) | Q(recipient=user))
+    accepted_friendships = friendships.filter(status=Friendship.FriendshipStatus.ACCEPTED)
+    return [model_to_dict(friendship) for friendship in accepted_friendships]
+
+class FriendshipViewSet(viewsets.ViewSet):
+    model = Friendship
+    queryset = Friendship.objects.all()
```
mureytasroc, line 397: Django has some really nice built-in functionality to auto-generate list methods (e.g. get) from the queryset you specify.
https://docs.djangoproject.com/en/4.1/topics/class-based-views/generic-display/#generic-views-of-objects
We might want to change this to:
```python
queryset = Friendship.objects.none()
def get_queryset(self):
    return Friendship.objects.filter(
        Q(sender=self.request.user) | Q(recipient=self.request.user),
        status__ne=Friendship.Status.REJECTED
    )
```
(including the line above `get_queryset` helps the auto-docs).
Then I think we can just delete the `get` method below. But we would have to make a Friendship [serializer](https://www.django-rest-framework.org/api-guide/serializers/) and include that in the ViewSet with e.g. `serializer_class = FriendshipSerializer`.

backend/courses/views.py
```diff
@@ -376,3 +384,122 @@ def get_queryset(self):
...
+            model_to_dict(friendship) for friendship in rejected_friendships
+        ]  # unsure if we should include this
+        res["message"] = "Friendships retrieved successfully."
+        return JsonResponse(res)
+                
+    def post(self, request):
+        res = {}
+        # send friendship request to passed in recipient
+        sender = request.user
+        recipient = get_object_or_404(User, id=request.recipient_id)
```
mureytasroc, line 425: If we want to get data from the body of the request, wouldn't it be `request.data.recipient_id`? Also I think we might want to call this `friend_id` so that both the sender and the recipient can use this method to send/accept friendships (discussed more below).

backend/courses/views.py
```diff
@@ -376,3 +384,122 @@ def get_queryset(self):
...
+
+        if not recipient:
+            res["message"] = "Recipient does not exist."
```
mureytasroc, line 428: I think this check is redundant due to the `get_object_or_404` call above -- the viewset will return 404 if it can't get the recipient.

backend/courses/views.py
```diff
@@ -376,3 +384,122 @@ def get_queryset(self):
...
+            return JsonResponse(res, status=400)
+
+        # if friend request already exists but was rejected, we can reinstate it:
+        friend_req = self.queryset.get(sender=sender, recipient=recipient, status=Friendship.FriendshipStatus.REJECTED)
```
mureytasroc, line 432: Might want to just get any existing friendship regardless of status (only making 1 trip to the db), and then case on the status once it's in-memory. We could do:
```python
existing_friendship = self.queryset.filter(
    Q(recipient=recipient) | Q(sender=recipient)
).first()
```

Using `first` is important because `get` will throw an exception if the object does not exist. `first` will just return `None`.

backend/courses/views.py
```diff
@@ -376,3 +384,122 @@ def get_queryset(self):
...
+
+        return JsonResponse(res, status=200)
+
+    def delete(self, request):
+        # either deletes a friendship or cancels a friendship request (depends on status)
+        res = {}
+        sender = request.user
+        recipient = get_object_or_404(User, id=request.recipient_id)
+
+        if not recipient:
```
mureytasroc, line 469: Same comment as above

backend/courses/views.py
```diff
@@ -376,3 +384,122 @@ def get_queryset(self):
...
+            res["message"] = "Recipient does not exist."
+            return JsonResponse(res, status=400)
+
+        # delete the friendship/request (see if it exists the other way around)
+        friendship = self.queryset.get(sender=sender, recipient=recipient)
```
mureytasroc, line 474: Might want to use the same code as suggested above to get any friendships (with any status / in any direction) between these two, and then case on the direction to determine whether the object should actually be deleted (i.e. when `existing_friendship.sender_id == request.user.id`, it should be deleted -- otherwise its status should be changed to `Friendship.Status.REJECTED`).

backend/courses/views.py
```diff
@@ -376,3 +384,122 @@ def get_queryset(self):
...
+            friendship = self.queryset.get(sender=recipient, recipient=sender)
+        if not friendship:
+            res["message"] = "Friendship/friendship request does not exist."
+        else:
+            friendship.delete()
+            res["message"] = "Friendship request removed."
+        return JsonResponse(res, status=200)
+    
+    # not sure what URL this method gets assigned to???
+    def handle(self, request):
```
mureytasroc, line 485: We can just handle this in the `POST` method - if user_a POSTs with friend=user_b, that should create the friendship with SENT status. Then if user_b POSTs with friend=user_a, that should find the existing friendship and change the status to ACCEPTED.

backend/courses/models.py
```diff
@@ -1,3 +1,4 @@
+from genericpath import exists
```
mureytasroc, line 1: Might want to move these models/views/urls under the plan folder, as they are more related to PCP specifically than courses. Courses is more for shared course-related functionality.

backend/tests/courses/test_friendship_models.py
```diff
@@ -0,0 +1,80 @@
+from django.contrib.auth.models import User
```
mureytasroc, line 1: Sorry ik this is an annoying comment, but usually we want to test code from the API interface, rather than testing specific implementation details (like the model). This makes the tests less brittle / less of "change indicators". It also provides more comprehensive testing, as lots of our interesting logic is in the viewset, not just the model. Issues with the model will surface if we properly test the code from the API level.

The schedules test are a good example of this:
https://github.com/pennlabs/penn-courses/blob/master/backend/tests/plan/test_schedule.py

You can actually send test requests e.g.:
```python
sender = User.objects.create_user(
    username="sender", email="sender@example.com", password="top_secret"
)
recipient = User.objects.create_user(
    username="recipient", email="recipient@example.com", password="top_secret"
)
self.client.login(username="sender", password="top_secret")
response_sent = self.client.post(
    "/api/plan/friendships/",
    json.dumps({"friend_id": recipient.id}),
    content_type="application/json",
)
# TODO: check response_sent for correctness, e.g. status code, content
recipient_client = APIClient()
recipient_client.login(username="recipient", password="top_secret")
response_accepted = recipient_client.put(
    "/api/plan/friendships/",
    json.dumps({"friend_id": sender.id}),
    content_type="application/json",
)
# TODO: check response_accepted for correctness, e.g. status code, content
```

backend/tests/plan/test_primary_schedules.py
```diff
@@ -0,0 +1,24 @@
...
+
+from alert.management.commands.recomputestats import recompute_precomputed_fields
+from alert.models import AddDropPeriod
+from courses.models import Friendship
+from tests.courses.util import create_mock_data
+
+class PrimaryScheduleTest(TestCase):
+    
+    def setUp(self):
+        # TODO: write test cases for the following cases
```
mureytasroc, line 16: This sounds like a good testing plan

backend/tests/courses/test_friendship_models.py
```diff
@@ -0,0 +1,80 @@
...
+
+        friendship = Friendship.objects.get(sender=u1, recipient=u2)
+        friendship.status = Friendship.FriendshipStatus.REJECTED
+        friendship.save()
+        self.assertTrue(Friendship.objects.filter(sender=u1, recipient=u2, status=Friendship.FriendshipStatus.REJECTED).exists())
+        self.assertFalse(Friendship.objects.filter(sender=u1, recipient=u2, status=Friendship.FriendshipStatus.SENT).exists())
+
+    
+    def test_friendship_route_logic(self):
+        # TODO: add functions to test the following friendship logic:
```
mureytasroc, line 67: Could just remove the tests above tbh, then make this file focused on testing from the API level

mureytasroc (commented):
Great progress, really clean code. Just some suggestions to make things more Django-friendly / utilize built-in functionality.

backend/courses/views.py
anshnagwekar: I'm planning on handling rejecting friendship requests in the post route. The delete route will only be to delete friendship requests or remove friends.

backend/courses/models.py
anshnagwekar: My reasoning for keeping all friendship-related stuff in courses/ was because we might use this functionality in PCR (and possibly PCA too) so it would be a good idea to organize it under courses/ for use in all courses products.

backend/courses/views.py
```diff
@@ -376,3 +384,122 @@ def get_queryset(self):
...
+    accepted_friendships = friendships.filter(status=Friendship.FriendshipStatus.ACCEPTED)
+    return [model_to_dict(friendship) for friendship in accepted_friendships]
+
+class FriendshipViewSet(viewsets.ViewSet):
+    model = Friendship
+    queryset = Friendship.objects.all()
+    http_method_names = ["get", "post", "delete"]
+    permission_classes = [IsAuthenticated]
+
+    # TODO: generate proper PcxAutoschema for this viewset
```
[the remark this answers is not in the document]
anshnagwekar: I actually wanted to ask you about this. How should I go about generating this PcxAutoSchema? I'm a little confused on what to write and how to structure it.

backend/courses/models.py
```diff
@@ -1419,3 +1420,69 @@ def create_or_update_user_profile(sender, instance, created, **kwargs):
...
+        user2 = User.objects.get(id=user2_id)
+        return (Friendship.objects.filter(sender=user1, recipient=user2, status="A").exists() or 
+                Friendship.objects.filter(sender=user2, recipient=user1, status="A").exists())
+
+    def save(self, *args, **kwargs):
+        if (self.status == self.FriendshipStatus.ACCEPTED):
+            self.accepted_at = timezone.now()
+        if (self.status == self.FriendshipStatus.REJECTED):
+            # any other logic that needs to happen when a friendship is rejected 
+            pass
```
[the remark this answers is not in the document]
anshnagwekar: I added some logic to reset the "sent_at" and "accepted_at" values to None when the request is rejected (in order to properly set the times in case the request has been sent again.

backend/courses/models.py
```diff
@@ -1419,3 +1420,69 @@ def create_or_update_user_profile(sender, instance, created, **kwargs):
...
+    )
+
+    def check_friendship(self, user1_id, user2_id):
+        """
+        Checks if two users are friends (lookup by user id)
+        """
+        user1 = User.objects.get(id=user1_id)
+        user2 = User.objects.get(id=user2_id)
+        return (Friendship.objects.filter(sender=user1, recipient=user2, status="A").exists() or 
+                Friendship.objects.filter(sender=user2, recipient=user1, status="A").exists())
```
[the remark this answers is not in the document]
mureytasroc: Ah sorry I didn't see this earlier. Actually if `sender` and `recipient` are fields in the model, you can filter with `sender_id` and `recipient_id` (Django implicitly adds fields with the `_id` suffix when you have related objects).

backend/courses/models.py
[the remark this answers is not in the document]
mureytasroc: Nice

backend/courses/models.py
mureytasroc: Hmm interesting, I guess that makes sense.

backend/plan/models.py
```diff
@@ -43,11 +46,54 @@ class Schedule(models.Model):
...
+        get_user_model(),
+        on_delete=models.CASCADE,
+        help_text="The person (user) to which the schedule belongs.",
+    )
+
+    schedule = models.ForeignKey(
+        Schedule,
+        on_delete=models.CASCADE,
+        help_text="The schedule that is the primary schedule for the user.",
+    )
```
mureytasroc, line 92: Let's make `schedule` a `OneToOneField` also with `related_name="primary_schedule"` -- we can use that `related_name` to check if a given schedule is primary.

backend/courses/views.py
```diff
@@ -376,3 +385,130 @@ def get_queryset(self):
             section__course__semester=get_current_semester(),
             in_add_drop_period=True,
         ).order_by("created_at")
+
+def get_accepted_friends(user):
+    """ Return user's accepted friends"""
+    Friendship.objects.filter(
+        Q(sender=user) | Q(recipient=user), 
+        status=Friendship.Status.ACCEPTED
+    ).values()
```
mureytasroc, line 394: Let's actually make this return a queryset (tweaked the usage in `courses/views` so that's more convenient). So in other words we can remove `.values()`.

backend/plan/views.py
```diff
@@ -176,6 +183,106 @@ def recommend_courses_view(request):
...
 
+class PrimaryScheduleViewSet(viewsets.ViewSet):
+    model = PrimarySchedule
+    queryset = PrimarySchedule.objects.none()
+    http_method_names = ["get", "post"]
+    permission_classes = [IsAuthenticated]
+    serializer_class = PrimaryScheduleSerializer
+
+    def get_queryset(self):
+        return PrimarySchedule.objects.all()
```
mureytasroc, line 194: `return PrimarySchedule.objects.filter(Q(user=self.request.user) | Q(user__in=get_accepted_friends(self.request.user)))`

backend/plan/views.py
```diff
@@ -176,6 +183,106 @@ def recommend_courses_view(request):
...
+                    403: "User does not have permission to use thiss schedule.",
+                }
+            },
+        },
+    )
+
+    def get(self, request):
+        res = {}
+        # return user primary schedule and all shared schedules with this user 
+        # if friend is passed in, then we return that specific friend's shared schedule
```
mureytasroc, line 217: Do we really need this functionality (filtering for a specific friend)? If not, we can just delete this whole method. Might be worth asking Andy/frontend. Less code is better

backend/courses/views.py
```diff
@@ -376,3 +385,130 @@ def get_queryset(self):
...
+
+class FriendshipViewSet(viewsets.ViewSet):
```
mureytasroc, line 396: Let's add a docstring to this class with some API documentation for each of the methods (GET, POST, DELETE) - I think this is necessary in this case because the flow is a bit unintuitive. This documentation will show up in our auto-docs under each of the new routes.

```python
"""
list: ...

create: ...

delete: ...
"""
```

backend/plan/views.py
```diff
@@ -176,6 +183,106 @@ def recommend_courses_view(request):
         status=status.HTTP_200_OK,
     )
 
+class PrimaryScheduleViewSet(viewsets.ViewSet):
```
mureytasroc, line 186: Let's add a docstring to this class with some API documentation for each of the methods (GET, PUT) - same reasoning as for friendships.

"""
list: ...

update: ...
"""

backend/plan/views.py
```diff
@@ -176,6 +183,106 @@ def recommend_courses_view(request):
...
+    model = PrimarySchedule
+    queryset = PrimarySchedule.objects.none()
+    http_method_names = ["get", "post"]
```
mureytasroc, line 189: Can we actually change post to put (idempotent) - make sure to change all POST references below if so

backend/plan/views.py
```diff
@@ -176,6 +183,106 @@ def recommend_courses_view(request):
...
+            res["message"] = "All primary schedules of friends retrieved."
+            res["schedules"] = friends
+            return JsonResponse(res)
+
+    def post(self, request):
+        res = {}
+        # verify that schedule exist and set it as the primary for the passed in user
+
+        user = request.user
+        schedule = get_object_or_404(Schedule, pk=request.data.schedule_id)
```
mureytasroc, line 262: `schedule = Schedule.objects.filter(person_id=user.id, id=request.data.schedule_id).first()`

backend/plan/views.py
```diff
@@ -176,6 +183,106 @@ def recommend_courses_view(request):
...
+        if not schedule:
+            res["message"] = "Schedule does not exist"
+            return JsonResponse(res, status=status.HTTP_STATUS_400_BAD_REQUEST)
+
+        # check syntax for this but ensure that the schedule belongs to the user
+        if schedule.person != user:
```
mureytasroc, line 268: Can delete this case if we make the change above.

backend/plan/views.py
```diff
@@ -176,6 +183,106 @@ def recommend_courses_view(request):
...
+            res["message"] = "Schedule does not belong to user"
+            return JsonResponse(res, status=status.HTTP_STATUS_403_FORBIDDEN)
+
+        # set the schedule as the primary schedule for the user (and return old shared schedule if exists)
+        primary_schedule_entry = self.queryset.get(person=user)
```
mureytasroc, line 273: `primary_schedule_entry = self.queryset.filter(user=user).first()`

backend/plan/views.py
```diff
@@ -176,6 +183,106 @@ def recommend_courses_view(request):
...
+        if primary_schedule_entry:
+            res['old_shared_schedule'] = model_to_dict(primary_schedule_entry.schedule)
+            res['new_shared_schedule'] = model_to_dict(schedule)
```
mureytasroc, line 276: I honestly don't think we need to respond with this information. Just setting `res["message"]` here is probably fine.

backend/plan/views.py
```diff
@@ -176,6 +183,106 @@ def recommend_courses_view(request):
...
+            primary_schedule_entry.schedule = schedule
+            primary_schedule_entry.save()
+            res["message"] = "Primary schedule successfully updated"
+        else:
+            self.queryset.create(person=user, schedule=schedule)
```
mureytasroc, line 281: `PrimarySchedule.objects.create(user=user, schedule=schedule)`

backend/plan/views.py
```diff
@@ -176,6 +183,106 @@ def recommend_courses_view(request):
...
+                    # no primary schedule associated with this friend
+                    friends.append( (query_friend, {}) ) 
+                    
+            res["message"] = "All primary schedules of friends retrieved."
+            res["schedules"] = friends
+            return JsonResponse(res)
+
+    def post(self, request):
+        res = {}
+        # verify that schedule exist and set it as the primary for the passed in user
```
mureytasroc, line 259: I think we can delete this comment

backend/plan/views.py
```diff
@@ -176,6 +183,106 @@ def recommend_courses_view(request):
...
+        primary_schedule_entry = self.queryset.get(person=user)
+        if primary_schedule_entry:
+            res['old_shared_schedule'] = model_to_dict(primary_schedule_entry.schedule)
+            res['new_shared_schedule'] = model_to_dict(schedule)
+            primary_schedule_entry.schedule = schedule
+            primary_schedule_entry.save()
+            res["message"] = "Primary schedule successfully updated"
+        else:
+            self.queryset.create(person=user, schedule=schedule)
+            res['new_shared_schedule'] = model_to_dict(schedule)
```
mureytasroc, line 282: Don't think `new_shared_schedule` is necessary to include in the response. The frontend already has that.

mureytasroc (commented):
Nice work. We're super close with this. Two high level comments:
- You might want to cut down on the amount of in-code comments that are just describing what the code does. Comments should be thought of as a shortcut - if you are commenting a complicated block of code to allow your reader to skip over that block (by just reading your comment), that's valuable. On the other hand if you have one line of comment describing one line of simple code, it's just adding bloat / taking up unnecessary space imo. Cleaner to just write good code and let it speak for yourself. The other situation where comments are valuable is when you have complicated invariants or implicit assumptions. Anything that can't be easily figured out from the code itself.
- I think sometimes you add a bit too much bloat to your JSON responses. Status messages are useful for error cases (to tell the user what went wrong), and sometimes it's useful to include some data in the content of your response to an action (e.g. I think returning the state of the friendship object after the POST friendship action has been executed is valuable). But in most cases if you are executing an action method (e.g. anything other than GET), you probably don't need to return a response json at all. The user already knows what they sent you (e.g. what resource they are POSTing, in the case of `PrimarySchedule`), and your status code should speak for itself.

backend/courses/views.py
```diff
@@ -376,3 +385,130 @@ def get_queryset(self):
...
+        ).first()
+
+        if not existing_friendship:
+            # friendship doesn't exist
+            res["message"] = "Friendship doesn't exist."
+            return JsonResponse(res, status=status.HTTP_404_NOT_FOUND)
+
+        # delete the friendship/request if it has already been accepted
+        if existing_friendship.status == Friendship.Status.ACCEPTED:
+            existing_friendship.delete()
```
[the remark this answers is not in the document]
anshnagwekar: I think this is the case when the friendship has already been accepted and now either the sender or recipient wants to remove the friendship. Thus, we would just delete the entry.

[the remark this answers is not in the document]
anshnagwekar: Unless we want to designate a "removed" friendship by setting its status to rejected? But that wouldn't really make sense.

backend/plan/models.py
```diff
@@ -43,11 +46,54 @@ class Schedule(models.Model):
...
+
+    class Meta:
+        # for now, we only allow one primary schedule per user
+        unique_together = (("person",),)
```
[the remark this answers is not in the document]
anshnagwekar: Yes but we also don't want multiple entries of the same user. But maybe that is now accomplished by changing it to a OneToOne field?

backend/plan/views.py
```diff
@@ -176,6 +183,106 @@ def recommend_courses_view(request):
...
+    model = PrimarySchedule
+    queryset = PrimarySchedule.objects.none()
+    http_method_names = ["get", "post"]
+    permission_classes = [IsAuthenticated]
+    serializer_class = PrimaryScheduleSerializer
+
+    def get_queryset(self):
+        return PrimarySchedule.objects.all()
+
+    # TODO: generate proper PcxAutoschema for this viewset
```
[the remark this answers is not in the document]
anshnagwekar: Completed earlier but forgot to commit the change. Please check now!

backend/courses/views.py
```diff
@@ -376,3 +385,130 @@ def get_queryset(self):
...
+        if not existing_friendship:
+            # create new friendship
+            friendship = Friendship(sender=sender, recipient=recipient, status=Friendship.Status.SENT)
+            friendship.save()
+            res = model_to_dict(friendship)
+            res["message"] = "Friendship request sent successfully."
+            return JsonResponse(res, status=200)
+
+        # reinstate old friendship request if it was previously rejected
+        if existing_friendship.status == Friendship.Status.REJECTED:
```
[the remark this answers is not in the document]
anshnagwekar: Sent a message to frontend / Andy.

backend/plan/views.py
anshnagwekar: I don't think so. If they need to filter by friend, they can just go through the primary schedules that they will already have loaded on the frontend and filter on "person" field there. Just going to delete the get method then.

backend/courses/models.py
```diff
@@ -1420,3 +1421,69 @@ def create_or_update_user_profile(sender, instance, created, **kwargs):
...
+        default=Status.SENT,
+    )
+
+    def are_friends(self, user1_id, user2_id):
+        """
+        Checks if two users are friends (lookup by user id)
+        """
+        return Friendship.objects.filter(
+            Q(sender_id=user1_id, recipient_id=user2_id, status="A")
+            | Q(sender_id=user2_id, recipient_id=user1_id, status="A")
```
mureytasroc, line 1464: Might want to use the Status enum instead of hardcoding "A"

backend/courses/views.py
```diff
@@ -376,3 +385,147 @@ def get_queryset(self):
...
+    )
+
+class FriendshipViewSet(viewsets.ViewSet):
+    """
+    list: Get a list of all friendships and friendship requests (sent and recieved) for the specified user.
+    Filter the list by status (accepted, sent) to distinguish between friendships and friendship requests.
+
+    create: Create a friendship between two users (sender and recipient). If a previous request does not exist between
+    the two friendships, then we create friendship request. If a previous request exists (where the recipient is the sender), 
+    then we accept the request. If the recipient of a request hits this route, then we accept the request and create the friendship.
```
mureytasroc, line 403: What's the difference between the last 2 cases? Aren't those the same?

backend/courses/views.py
```diff
@@ -376,3 +385,147 @@ def get_queryset(self):
...
+    schema = PcxAutoSchema(
+        response_codes={
+            reverse_func("friendships"): {
+                "GET": {
+                    200: "Friendships retrieved successfully.",
+                },
+                "POST": {
+                    200: "Friendship request handled successfully.",
+                    409: "Friendship request already exists",
+                    400: "Bad request.",
```
mureytasroc, line 424: Do we ever actually return 400? If not we can remove this.

backend/courses/views.py
```diff
@@ -376,3 +385,147 @@ def get_queryset(self):
...
+    delete: Delete a friendship between two users (sender and recipient). If there exists only a friendship request between two users, then 
+    we either delete the friendship request if the sender hits the route, or we reject the request if the recipient hits this route.
+    """
+    model = Friendship
+    http_method_names = ["get", "post", "delete"]
+    permission_classes = [IsAuthenticated]
+    queryset = Friendship.objects.none()
+    serializer_class = FrienshipSerializer
+
+    # Is this PcxAutoSchema correct given the responses in each view function below?
```
mureytasroc, line 414: I think so - just made 1 comment

backend/courses/views.py
```diff
@@ -376,3 +385,147 @@ def get_queryset(self):
...
+        recipient = get_object_or_404(User, id=request.GET.get("friend_id"))
+
+        existing_friendship = self.queryset.filter(
+            Q(recipient=recipient) | Q(sender=recipient)
+        ).first()
+
+        if not existing_friendship:
+            friendship = Friendship(sender=sender, recipient=recipient, status=Friendship.Status.SENT)
+            friendship.save()
+            res = model_to_dict(friendship)
```
mureytasroc, line 482: Should use `FriendshipSerializer` rather than `model_to_dict` (this is typically true, for security reasons).

backend/courses/views.py
```diff
@@ -376,3 +385,147 @@ def get_queryset(self):
...
+            return JsonResponse(res, status=status.HTTP_200_OK)
+        elif existing_friendship.status == Friendship.Status.REJECTED:
+            existing_friendship.status = Friendship.Status.SENT
+            existing_friendship.sender = sender
+            existing_friendship.recipient = recipient
+            existing_friendship.save()
+            res = model_to_dict(existing_friendship)
```
mureytasroc, line 489: `FriendshipSerializer`

backend/courses/views.py
```diff
@@ -376,3 +385,147 @@ def get_queryset(self):
...
+            return JsonResponse(res, status=status.HTTP_200_OK)
+        elif existing_friendship.status == Friendship.Status.SENT:
+            if existing_friendship.sender == sender:
+                res["message"] = "Friendship request already sent."
+                return JsonResponse(res, status=status.HTTP_409_CONFLICT)
+            elif existing_friendship.recipient == sender:
+                existing_friendship.status = Friendship.Status.ACCEPTED
+                existing_friendship.save()
+                res = model_to_dict(existing_friendship)
```
mureytasroc, line 498: `FriendshipSerializer`

backend/courses/views.py
mureytasroc, line 489: `FriendshipSerializer`

backend/courses/views.py
```diff
@@ -376,3 +385,147 @@ def get_queryset(self):
...
+        existing_friendship = self.queryset.filter(
+            Q(recipient=recipient) | Q(sender=recipient)
+        ).first()
+
+        if not existing_friendship:
+            friendship = Friendship(sender=sender, recipient=recipient, status=Friendship.Status.SENT)
+            friendship.save()
+            res = model_to_dict(friendship)
+            return JsonResponse(res, status=status.HTTP_200_OK)
+        elif existing_friendship.status == Friendship.Status.REJECTED:
```
mureytasroc, line 484: Double check that this is desired behavior with Andy

backend/plan/models.py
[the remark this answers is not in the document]
mureytasroc: Oh my bad, ur right

backend/plan/serializers.py
```diff
@@ -16,3 +17,12 @@ class Meta:
         model = Schedule
         exclude = ["person"]
         extra_kwargs = {"semester": {"required": False}}
+
+class PrimaryScheduleSerializer(serializers.ModelSerializers):
+    schedule = ScheduleSerializer(
+        read_only=False, help_text="The primary schedule.", required=True
```
mureytasroc, line 23: `read_only=True, required=False` ?

mureytasroc (commented):
LGTM so far

PR #446, opened 2022-12-30
Advanced Registration View

alex-budko:
Added linting changes to the advanced registration view

### DIFF
--- a/backend/plan/models.py
+++ b/backend/plan/models.py
@@ -43,11 +43,23 @@ class Schedule(models.Model):
         """
         ),
     )
+
+    advanced_registration = models.BooleanField(
+        blank=False,
+        default=False,
+        help_text=dedent(
+            """
+            A label to denote whether the user's schedule was made for advanced registration.
+            """
+        ),
+    )
+
     created_at = models.DateTimeField(auto_now_add=True)
     updated_at = models.DateTimeField(auto_now=True)
 
     class Meta:
         unique_together = (("name", "semester", "person"),)
+        unique_together = (("name", "semester", "person", "advanced_registration"), ("semester", "person", "advanced_registration"))
 
     def __str__(self):
         return "User: %s, Schedule ID: %s" % (self.person, self.id)
--- a/backend/plan/views.py
+++ b/backend/plan/views.py
@@ -77,6 +77,92 @@ from plan.serializers import ScheduleSerializer
         },
     )
 )
+
+# PcxAutoSchema for the advanced registration view.
+@api_view(["POST"])
+@schema(
+    PcxAutoSchema(
+        response_codes={
+            reverse_func("advanced-registration-schedule"): {
+                "POST": {
+                    200: "[DESCRIBE_RESPONSE_SCHEMA]Response returned successfully.",
+                    201: "[UNDOCUMENTED]",
+                    400: "Invalid advanced_reg_codes, advanced_reg_sections, schedule_user, or advanced_reg_schedule_name (see response).",
+                }
+            }
+        },
+        override_request_schema={
+            reverse_func("advanced-registration-schedule"): {
+                "POST": {
+                    "type": "object",
+                    "properties": {
+                        "advanced_reg_codes": {
+                            "type": "array",
+                            "description": (
+                                "An array of courses the user is currently planning to "
+                                "take, each specified by its string full code, of the form "
+                                "e.g. CIS-120-000."
+                            ),
+                            "items": {"type": "string"},
+                        },
+                        "advanced_reg_schedule_name": {
+                            "type": "string",
+                            "description": (
+                                "Name of the advanced registration schedule as "
+                                "Specified by the user "
+                            ),
+                        },
+                    },
+                }
+            }
+        },
+        override_response_schema={
+            reverse_func("advanced-registration-schedule"): {
+                "POST": {
+                    200: {"type": "object"}
+                }
+            }
+        },
+    )
+)
+
+@permission_classes([IsAuthenticated])
+def advanced_registration_schedule_view(request):
+    """
+    This route will takes in advanced registration codes and a schedule's name
+    to compose an advanced registration schedule and save it to the database. This view
+    also returns the created schedule in the response.
+    """
+
+    schedule_user = request.user
+    advanced_reg_codes = request.data.get("advanced_reg_codes", None)
+    advanced_reg_codes = advanced_reg_codes if advanced_reg_codes is not None else []
+
+    advanced_reg_schedule_name = request.data.get("advanced_reg_schedule_name")
+    semester = request.data.get("semester", get_current_semester())
+
+    try:
+        advanced_reg_sections = []
+        for advanced_reg_code in advanced_reg_codes:
+            advanced_reg_sections.append(Section.objects.filter(code=advanced_reg_code))
+
+        if len(advanced_reg_sections) == 0:
+            return Response(
+                f"Empty Advanced Registration Schedule",
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+
+        advanced_reg_schedule = Schedule.objects.create(person=schedule_user, sections=advanced_reg_sections,
+            semester=semester, name=advanced_reg_schedule_name, advanced_registration=True)
+        advanced_reg_schedule.save()
+
+    except ValueError as e:
+        return Response(
+            str(e),
+            status=status.HTTP_400_BAD_REQUEST,
+        )
+    return Response({"message": "success", "advanced_reg_schedule": advanced_reg_schedule}, status=status.HTTP_200_OK)
+    
 @permission_classes([IsAuthenticated])
 def recommend_courses_view(request):
     """
--- a/backend/plan/urls.py
+++ b/backend/plan/urls.py
@@ -2,8 +2,7 @@ from django.urls import include, path
 from django.views.generic import TemplateView
 from rest_framework import routers
 
-from plan.views import ScheduleViewSet, recommend_courses_view
-
+from plan.views import ScheduleViewSet, recommend_courses_view, advanced_registration_schedule_view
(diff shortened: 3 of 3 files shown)


alex-budko:
Unlinted copy of the PR for the advanced registration view

backend/plan/models.py
```diff
@@ -43,11 +43,23 @@ class Schedule(models.Model):
...
+            """
+        ),
+    )
+
     created_at = models.DateTimeField(auto_now_add=True)
     updated_at = models.DateTimeField(auto_now=True)
 
     class Meta:
         unique_together = (("name", "semester", "person"),)
+        unique_together = (("name", "semester", "person", "advanced_registration"), ("semester", "person", "advanced_registration"))
```
AaDalal, line 62: Each tuple is `and`ed together -- so both tuples have to be unique, otherwise everything fails. The first tuple ends up being redundant because it just a subset of the cases covered by the second tuple.

I would fix this by deleting the second constraint. When advanced registration is False (all of the cases prior to this PR, we have the original constraint of `("name", "semester", "person"))`. When `advanced_registration` is true, we should only have 1 schedule per semester so the `("name", "semester", "person")` constraint should always succeed.

backend/plan/views.py
```diff
@@ -77,6 +77,92 @@
...
+        override_response_schema={
+            reverse_func("advanced-registration-schedule"): {
+                "POST": {
+                    200: {"type": "object"}
+                }
+            }
+        },
+    )
+)
+
```
AaDalal, line 128: I would delete this space. Have you tried running the documentation locally to see how the auto docs look?

backend/plan/views.py
```diff
@@ -77,6 +77,92 @@
...
+# PcxAutoSchema for the advanced registration view.
+@api_view(["POST"])
+@schema(
+    PcxAutoSchema(
+        response_codes={
+            reverse_func("advanced-registration-schedule"): {
+                "POST": {
+                    200: "[DESCRIBE_RESPONSE_SCHEMA]Response returned successfully.",
+                    201: "[UNDOCUMENTED]",
```
AaDalal, line 89: Is there ever a case where 201 is returned?

backend/plan/views.py
```diff
@@ -77,6 +77,92 @@
...
+                                "take, each specified by its string full code, of the form "
+                                "e.g. CIS-120-000."
+                            ),
+                            "items": {"type": "string"},
+                        },
+                        "advanced_reg_schedule_name": {
+                            "type": "string",
+                            "description": (
+                                "Name of the advanced registration schedule as "
+                                "Specified by the user "
```
AaDalal, line 112: Super tiny nit: `specified` and period at the end

backend/plan/views.py
```diff
@@ -77,6 +77,92 @@
...
+                        "advanced_reg_codes": {
+                            "type": "array",
+                            "description": (
+                                "An array of courses the user is currently planning to "
+                                "take, each specified by its string full code, of the form "
+                                "e.g. CIS-120-000."
+                            ),
+                            "items": {"type": "string"},
+                        },
+                        "advanced_reg_schedule_name": {
```
AaDalal, line 108: Assuming that we only allow 1 adv registration schedule per semester per user, does this make sense to include? Should the name just be something like "advanced registration"

backend/plan/views.py
```diff
@@ -77,6 +77,92 @@
...
+                }
+            }
+        },
+    )
+)
+
+@permission_classes([IsAuthenticated])
+def advanced_registration_schedule_view(request):
+    """
+    This route will takes in advanced registration codes and a schedule's name
```
AaDalal, line 132: Maybe clarify what codes we're referring to here.

backend/plan/views.py
```diff
@@ -77,6 +77,92 @@
...
+    to compose an advanced registration schedule and save it to the database. This view
+    also returns the created schedule in the response.
```
AaDalal, line 134: I would elaborate more on why this would return 400. For example (from `recommend_courses_view(...)`)
```
    If successful, this route will return a list of recommended courses, with the same schema
    as the List Courses route, starting with the most relevant course. The number of
    recommended courses returned can be specified using the n_recommendations attribute in the
    request body, but if this attribute is omitted, the default will be 5.
    If n_recommendations is not an integer, or is <=0, a 400 will be returned.
    If curr_courses contains repeated courses or invalid courses or non-current courses, a
    400 will be returned.
    If past_courses contains repeated courses or invalid courses, a 400 will be returned.
    If curr_courses and past_courses contain overlapping courses, a 400 will be returned
```

backend/plan/views.py
```diff
@@ -77,6 +77,92 @@
...
+    """
+
+    schedule_user = request.user
+    advanced_reg_codes = request.data.get("advanced_reg_codes", None)
+    advanced_reg_codes = advanced_reg_codes if advanced_reg_codes is not None else []
```
AaDalal, line 139: You could simplify the two lines above as just `advanced_reg_codes = request.data.get("advanced_reg_codes", [])`

backend/plan/views.py
```diff
@@ -77,6 +77,92 @@
...
+
+    advanced_reg_schedule_name = request.data.get("advanced_reg_schedule_name")
```
AaDalal, line 141: What if this is not included in the request?

backend/plan/views.py
```diff
@@ -77,6 +77,92 @@
...
+                    400: "Invalid advanced_reg_codes, advanced_reg_sections, schedule_user, or advanced_reg_schedule_name (see response).",
+                }
+            }
+        },
+        override_request_schema={
+            reverse_func("advanced-registration-schedule"): {
+                "POST": {
+                    "type": "object",
+                    "properties": {
+                        "advanced_reg_codes": {
```
AaDalal, line 99: Where is `semester`?

backend/plan/views.py
```diff
@@ -77,6 +77,92 @@
...
+                            "type": "array",
+                            "description": (
+                                "An array of courses the user is currently planning to "
+                                "take, each specified by its string full code, of the form "
+                                "e.g. CIS-120-000."
```
AaDalal, line 104: Nit: don't need the "e.g." on this line

backend/plan/views.py
```diff
@@ -77,6 +77,92 @@
...
+    semester = request.data.get("semester", get_current_semester())
+
+    try:
+        advanced_reg_sections = []
+        for advanced_reg_code in advanced_reg_codes:
+            advanced_reg_sections.append(Section.objects.filter(code=advanced_reg_code))
+
+        if len(advanced_reg_sections) == 0:
+            return Response(
+                f"Empty Advanced Registration Schedule",
```
AaDalal, line 151: Nit: unnecessary formatted string (remove the leading f)

backend/plan/views.py
```diff
@@ -77,6 +77,92 @@
...
+    advanced_reg_codes = request.data.get("advanced_reg_codes", None)
+    advanced_reg_codes = advanced_reg_codes if advanced_reg_codes is not None else []
+
+    advanced_reg_schedule_name = request.data.get("advanced_reg_schedule_name")
+    semester = request.data.get("semester", get_current_semester())
+
+    try:
+        advanced_reg_sections = []
+        for advanced_reg_code in advanced_reg_codes:
+            advanced_reg_sections.append(Section.objects.filter(code=advanced_reg_code))
```
AaDalal, line 147: can simplify this for loop into `Section.objects.filter(code__in=advanced_reg_codes)`

backend/plan/views.py
AaDalal, line 151: Maybe an error code would be `No valid sections provided` (or something along those lines)

backend/plan/views.py
```diff
@@ -77,6 +77,92 @@
...
+            return Response(
+                f"Empty Advanced Registration Schedule",
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+
+        advanced_reg_schedule = Schedule.objects.create(person=schedule_user, sections=advanced_reg_sections,
+            semester=semester, name=advanced_reg_schedule_name, advanced_registration=True)
+        advanced_reg_schedule.save()
+
+    except ValueError as e:
```
AaDalal, line 159: What causes this `ValueError`? It's generally better practice to avoid try-except clauses and instead ensure your inputs are correct before using them.

backend/plan/views.py
```diff
@@ -77,6 +77,92 @@
...
+        return Response(
+            str(e),
+            status=status.HTTP_400_BAD_REQUEST,
+        )
+    return Response({"message": "success", "advanced_reg_schedule": advanced_reg_schedule}, status=status.HTTP_200_OK)
```
AaDalal, line 164: Does this response look correct? I'm actually unsure whether returning a django model instance automatically serializes it.

AaDalal (changes requested):
Good work here, but there are a few things to handle before we merge:
- Do we want to allow more than 1 advanced registration schedule per user per semester? If so, why? If not, we need to preserve that invariant, probably by overriding the `save` method of `Schedule`.
- How can we make sure that the inputs to the view are properly validated?

Couple of general notes:
- **IMPORTANT**: whenever you make changes to a model, you should run `makemigrations` and `migrate` and commit the migrations file that produces. I don't see the migration here, which could be problematic!
- **IMPORTANT**: you need to test your code using the unit testing framework (see the `tests` folder)
- make the description of the pull request more descriptive (see some of our previous PRs, particular @mureytasroc's for inspo). 
- The title can similarly more clearly describe what this feature adds.

backend/plan/views.py
AaDalal: One other important note: we only want the `Section` objects from the semester in question, so I think it should be `Section.objects.filter(code__in=advanced_reg_codes, semester=semester)`

PR #455, opened 2023-01-07
Ics-download-feature

alex-budko:
iCalendar download feature

### DIFF
--- a/frontend/plan/actions/index.js
+++ b/frontend/plan/actions/index.js
@@ -25,6 +25,7 @@ export const REMOVE_SCHED_ITEM = "REMOVE_SCHED_ITEM";
 export const DELETE_SCHEDULE = "DELETE_SCHEDULE";
 export const RENAME_SCHEDULE = "RENAME_SCHEDULE";
 export const DUPLICATE_SCHEDULE = "DUPLICATE_SCHEDULE";
+export const DOWNLOAD_SCHEDULE = "DOWNLOAD_SCHEDULE";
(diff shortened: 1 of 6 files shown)


rohangpta:
Looks like a great initial pass on the feature! A couple notes on this: 

- I think adding an RRule is probably a good idea. I envision the use case of this product as being something that people use to export their schedules and add to their respective GCals.
- In that vein, we should probably provide an API endpoint which returns and ICS, as opposed to having people download a `.ics` file. Please take a look at the Export functionality on this webpage: https://pennclubs.com/events. We would essentially provide an API endpoint which returns the ICS formatted calendar for a particular schedule. Folks can then add this to their Google Calendar.  Here is the [view](https://github.com/pennlabs/penn-clubs/blob/82dc1edd1bcdf25b6467c9ed409939cf82b120ea/backend/clubs/views.py#L3262) that Penn Clubs uses.

This has the primary benefit of auto updating on changes to the API result. So, if people update their primary schedule on PCP, then their GCal will also update; this means that this will be a **one time** calendar addition process for the rest of their Penn career (no repetition for add/drop/across semesters) -- all people need to do is update their "main" schedule on PCP.

I am happy to collaborate on this, so let me know what you think. This seems like a happy medium before we're able to actually auth into Path@Penn and get people's schedules (might not be the easiest thing in the world to achieve, from an authentication standpoint).

alex-budko:
I agree with all the points that you have made, Rohan! I think it would be more feasible for a better user experience to provide an API endpoint that returns an ICS, as opposed to having people download a .ics file. I will start working on implementing this pivot right away!

PR #461, opened 2023-01-29
Is open

susanzzz123:
is-opne
**Done**

- Added toggle button to filter classes that are currently open
- Refractored the ButtonFilter component to accommodate for an additional button that takes in a different argument type


frontend/plan/components/search/ButtonFilter.tsx
```diff
@@ -11,30 +11,30 @@ import {
...
 
-export function FilterButton<
-    F extends { [P in K]: number },
+export function ButtonFilter<
+    F extends { [P in K]: number | boolean },
     K extends keyof F>({
     title,
     filterData,
     defaultFilter,
     clearFilter,
```
andyjiang3, line 32: Do we need `defaultFilter`? It doesnt seem to be used

frontend/plan/components/search/ButtonFilter.tsx
```diff
@@ -58,7 +68,7 @@ export function FilterButton<
                 <DropdownFilterButton
                     defaultData={!isActive}
                     aria-haspopup="true"
-                    aria-controls="dropdown-menu"
+                    // aria-controls="dropdown-menu"
                     onClick={toggleButton}
```
andyjiang3, line 72: Lets remove both `aria-controls` and `aria-haspopup`

frontend/plan/components/search/ButtonFilter.tsx
```diff
@@ -43,11 +43,21 @@ export function FilterButton<
...
+            } 
+            if (buttonProperty === "is_open") {
+                updateButtonFilter(true);
+                startSearch({
+                    ...filterData,
+                    [buttonProperty]: true,
+                });
+            } 
+            
         }
```
andyjiang3, line 61: Can we generalize the value so we don't have to handle different buttonProperty, maybe lets change is_open to take number like 0 or -1 for false and 1 for true. This way we don't need both if statement and remove `having number | boolean` type

andyjiang3 (changes requested):
Great job on this! Just some generalization suggestions.

frontend/plan/types.ts
```diff
@@ -185,7 +185,8 @@ export interface FilterData {
         U: boolean;
     };
     time: [number, number];
-    fit_schedule: number;
+    "schedule-fit": number;
+    is_open: number;
```
andyjiang3, line 189: Is there a reason why` schedule-fit` is a string here?

frontend/plan/components/search/SearchBar.tsx
```diff
@@ -72,7 +72,7 @@ interface SearchBarProps {
     storeLoaded: boolean;
     setShowLoginModal: React.Dispatch<React.SetStateAction<boolean>>;
     activeSchedule: number;
-    updateButtonFilter: (field: string) => (value: number) => void;
+    updateButtonFilter: (field: string) => (value: number | boolean) => void;
 }
```
andyjiang3, line 76: Is there a case when `value` can be boolean?

andyjiang3 (changes requested):
Looks great, just 2 very small comments before we're good to merge! Also looks like one of the backend CI didn't pass, I think it might be caused by one of the commits in master?

andyjiang3 (approved):
LGTM!

PR #480, opened 2023-03-26
Changed font size so that the course codes fit on the block more nicely.

kboonstra:
Font size to 0.9em.


rohangpta:
LGTM Kepler!

PR #501, opened 2023-08-12
Alex-budko/issue491

alex-budko:
Fixed the UI for PennCoursePlan to show a download icon next to each calendar, implying that we can independently download/import any one of our named schedules:


backend/Pipfile
```diff
@@ -60,6 +60,9 @@ scikit-learn = "*"
 pandas = "*"
 python-dateutil = "*"
 docutils = "*"
+ics = "*"
+psycopg2-binary = "*"
```
AaDalal, line 64: could we add this under dev packages at the very least?

backend/courses/migrations/0054_userprofile_uuid_secret_and_more.py
```diff
@@ -0,0 +1,27 @@
+# Generated by Django 4.1.5 on 2023-01-15 16:39
```
AaDalal, line 1: can we merge this migration with the migrations on master?

backend/plan/urls.py
```diff
@@ -1,15 +1,15 @@
 from django.urls import include, path
 from django.views.generic import TemplateView
-from rest_framework import routers
+from rest_framework_nested import routers
```
AaDalal, line 3: Wait why do we use rest_framework_nested here?

backend/plan/views.py
```diff
@@ -399,3 +406,101 @@ def get_queryset(self):
...
+            reverse_func("calendar-view", args=["schedule_pk"]): {
+                "GET": {
+                    200: "Schedule exported successfully",
+                },
+            },
+        },
+    )
+
+    def get(self, *args, **kwargs):
+        """
```
AaDalal, line 423: we usually document this information in PcxAutoSchema -- see [here](https://github.com/pennlabs/penn-courses/blob/88de7109d2d2f6898719b1685baa0405131d838c/backend/PennCourses/docs_settings.py#L208) for an example.

backend/plan/views.py
```diff
@@ -399,3 +406,101 @@ def get_queryset(self):
...
+                        schema:
+                            type: string
+        ---
+        """
+        schedule_pk = kwargs["schedule_pk"]
+
+        schedule = (
+            Schedule.objects.filter(pk=schedule_pk)
+            .prefetch_related("sections", "sections__meetings")
+            .first()
```
AaDalal, line 440: nit: we should use `.get()` I think since we only expect one here

backend/plan/views.py
```diff
@@ -399,3 +406,101 @@ def get_queryset(self):
...
+        )
+
+        if not schedule:
+            return Response({"detail": "Invalid schedule"}, status=status.HTTP_403_FORBIDDEN)
+
+        day_mapping = {"M": "MO", "T": "TU", "W": "WE", "R": "TH", "F": "FR"}
```
AaDalal, line 446: probably better practice to just declare this at the top level (not inside the function) and make it all caps

backend/plan/views.py
```diff
@@ -399,3 +406,101 @@ def get_queryset(self):
...
+            e = ICSEvent()
+            e.name = section.full_code
+            e.created = timezone.now()
+
+            days = []
+            for meeting in section.meetings.all():
+                days.append(day_mapping[meeting.day])
+            first_meeting = list(section.meetings.all())[0]
+
+            start_time = str(Meeting.int_to_time(first_meeting.start))
```
AaDalal, line 461: Just a general question -- is there a reason we do this from the backend instead of the frontent?

backend/plan/views.py
```diff
@@ -399,3 +406,101 @@ def get_queryset(self):
...
+
+        for section in schedule.sections.all():
+            e = ICSEvent()
+            e.name = section.full_code
+            e.created = timezone.now()
+
+            days = []
+            for meeting in section.meetings.all():
+                days.append(day_mapping[meeting.day])
+            first_meeting = list(section.meetings.all())[0]
```
AaDalal, line 459: is any sorted order intended here?

frontend/package.json
```diff
@@ -20,5 +20,7 @@
     "hooks": {
       "pre-commit": "lerna run --concurrency 1 precommit"
     }
+  },
+  "dependencies": {
```
AaDalal, line 24: do we need to have a dependencies obj?

frontend/plan/components/modals/CalendarModal.tsx
```diff
@@ -0,0 +1,150 @@
...
+                    >
+                        Copy
+                    </a>
+                </Col>
+            </Row>
+
+            <hr />
+
+            <Row className="columns has-text-centered">
+                <div className="column">
```
AaDalal, line 104: nit but I think we use 2 space indentation -- you also may need to lint the frontend

frontend/plan/components/modals/CalendarModal.tsx
```diff
@@ -0,0 +1,150 @@
...
+                        onClick={(e) => (e.target as HTMLInputElement).select()}
+                    />
+                </Col>
+                <Col className="col control">
+                    <a
+                        className="button is-info"
+                        onClick={async () => {
+                            try {
+                                await navigator.clipboard.writeText(url);
+                                toast.info("Copied to clipboard!");
```
AaDalal, line 88: how does this toast look? We should make sure it's visually consistent

AaDalal (changes requested):
Left a few comments -- here's a summary
1. We should lint front + backend
2. Let's make sure this is visually consistent. Can you send a screenshot of how it looks in the penn courses channel on slack?
3. There are some weird dependency adds + dev set up changes on the backend. Let's remove or chat about those.

backend/plan/views.py
alex-budko: When I first built the feature I had all the logic done on the frontend, but Rohan suggested that logic be moved to the backend to simplify the frontend code and make it look more visually appealing

AaDalal: Ok yeah I think the link is maybe more intuitive.

AaDalal:
One other thing I thought of -- it might be nice to have the calendar show up with a default name when you add the URL to google (not sure if it's possible or how much work it is -- just a thought)

backend/README.md
```diff
@@ -78,20 +76,20 @@ NOTE: when using `pipenv`, environment variables are only refreshed when you exi
...
+    - To alias this command, run `echo "alias courses-compose='cd "$PWD"; docker-compose up'" >> ~/.zshrc; source ~/.zshrc` (replacing `~/.zshrc` with `~/.bashrc` or whatever configuration file your shell uses, if you don't use zsh).
+    - This will spin up a container from which you can run the server (with all required packages preinstalled). 
+3. In a separate terminal (from any directory), run `docker exec -it backend-development-1 /bin/bash` to open a shell in the container (if this says "no such container", run `docker container ls` and use the name of whatever container most closely matches the `backend_development` image). Just like exiting a Pipenv shell, you can exit the container by pressing `Ctrl+D` (which sends the "end of transmission" / EOF character).
```
andyjiang3, line 88: The numbering is off here, should be 2

frontend/plan/components/modals/CalendarModal.tsx
```diff
@@ -1,137 +1,156 @@
...
+    margin: 0 5px;
+    border-radius: 5px;
 `;
 
-const BigText = styled.p`
-    font-size: 14px;
-    line-height: 1.5;
+const ICSButton = styled(Button)`
+    background-color: #07aced;
+`;
```
andyjiang3, line 36: Is there a reason why the ICS URL label is a button? We should change this to a text

frontend/plan/components/modals/CalendarModal.tsx
```diff
@@ -1,137 +1,156 @@
...
+        props.show ? "visible" : "hidden"};
+`;
+
+const Toast: React.FC<ToastProps> = ({ message, show }) => {
+    return <ToastWrapper show={show}>{message}</ToastWrapper>;
+};
+
+const CalendarModal = ({
+    schedulePk,
+    clickedOnSchedule,
```
andyjiang3, line 91: Do we need `schedulePk` anymore? Seems like we can just use `chickedOnSchedule`

frontend/plan/components/modals/CalendarModal.tsx
```diff
@@ -1,137 +1,156 @@
...
-                        <a
-                            href="https://support.apple.com/guide/calendar/subscribe-to-calendars-icl1022/mac"
-                            target="_blank"
-                            rel="noreferrer noopener"
-                        >
-                            Check out this guide!
-                        </a>
-                    </BigText>
-                </div>
-            </Row>
```
andyjiang3, line 134: Is there a reason why we're removing this? Seems like it would be helpful to include for people that don't know how to use ICS URL

andyjiang3 (changes requested):
Left some small comments. Can we also add some margin between the description text and the ics url and make the ics url label width smaller so that the url is shown more

Screenshot 2023-09-04 at 6 54 16 PM

PR #503, opened 2023-08-22
PCR: limit fields aggregated on an instructor's courses

AaDalal:
This pull request limits the fields aggregated for each course an instructor teaches to those displayed by the frontend, namely:
1. `course_quality`
2. `instructor_quality`
3. `difficulty`
4. `work_required`
It limits the fields for both `average` and `recent` aggregations.

### DIFF
--- a/backend/review/annotations.py
+++ b/backend/review/annotations.py
@@ -258,7 +258,7 @@ def annotate_with_matching_reviews(
     )
 
 
-def annotate_average_and_recent(qs, match_review_on, match_section_on, extra_metrics=True):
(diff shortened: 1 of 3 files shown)


rohangpta:
Is this not the default? Or is this PR just stating that explicitly

AaDalal:
> Is this not the default? Or is this PR just stating that explicitly

No not the default, this is to speed up the particular part of the query that's slow

rohangpta:
Great thanks for clarifying. CI is passing, LGTM

rohangpta (approved):

PR #504, opened 2023-08-23
PCR optimize instructor's course query

AaDalal:
> NOTE: this is not currently working! This is a useful baseline but it is only a starting point.
> Use the `USE_NEW_QUERIES` toggle in `review.views` to switch between using the new (not working) queries and the old (working but slow) queries. After you flip this toggle you can run the server and visit a URL like `http://127.0.0.1:8000/api/review/instructor/14021` to see what the new and old queries return, respectively.

This PR aims to optimize the query that aggregates reviews for an instructor's courses.

### DIFF
--- a/backend/review/views.py
+++ b/backend/review/views.py
@@ -1,7 +1,7 @@
 from collections import defaultdict
 
 from dateutil.tz import gettz
-from django.db.models import F, Max, OuterRef, Q, Subquery, Value
+from django.db.models import F, Max, OuterRef, Q, Subquery, Value, Avg, FloatField
 from django.http import Http404
 from django.shortcuts import get_object_or_404
 from rest_framework.decorators import api_view, permission_classes, schema
@@ -26,7 +26,7 @@ from review.documentation import (
     instructor_for_course_reviews_response_schema,
     instructor_reviews_response_schema,
 )
-from review.models import ALL_FIELD_SLUGS, Review
+from review.models import ALL_FIELD_SLUGS, Review, ReviewBit
 from review.util import (
     aggregate_reviews,
     avg_and_recent_demand_plots,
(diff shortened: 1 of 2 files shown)


backend/review/views.py
```diff
@@ -372,6 +373,8 @@ def check_instructor_id(instructor_id):
     "difficulty",
 ]
 
+# TODO: remove before merging into prod
```
rohangpta, line 376: Commenting here as reminder

backend/review/views.py
```diff
@@ -446,6 +490,7 @@ def instructor_reviews(request, instructor_id):
     courses_res = dict()
     max_sem = dict()
     for r in courses.values():
+        print(r)
```
rohangpta, line 493: rm

rohangpta:
Is the idea behind this PR to remove some of the python logic that gets max sem and port over to SQL? 

Why do we see these performance issues to begin with? We are caching pages for 1 month as per `backend/review/urls.py` and requests after that should be quite free. Should we try to also warm the cache somehow?

Have you spoken to your friend about scraping our data? I worry we're seeing some interference here

AaDalal:
> Is the idea behind this PR to remove some of the python logic that gets max sem and port over to SQL?
> 
> Why do we see these performance issues to begin with? We are caching pages for 1 month as per `backend/review/urls.py` and requests after that should be quite free. Should we try to also warm the cache somehow?
> 
> Have you spoken to your friend about scraping our data? I worry we're seeing some interference here

Yeah -- he's not doing any scraping

AaDalal:
> Is the idea behind this PR to remove some of the python logic that gets max sem and port over to SQL?
> 
> Why do we see these performance issues to begin with? We are caching pages for 1 month as per `backend/review/urls.py` and requests after that should be quite free. Should we try to also warm the cache somehow?
> 
> Have you spoken to your friend about scraping our data? I worry we're seeing some interference here

Not it's changing the output query logic (e.g., you can try flipping the toggle and looking at the generated SQL queries. The ones output by the current query logic are really complex.

PR #510, opened 2023-09-18
simple split topics script

AaDalal:
This pull requests adds a simple script to break up topics and re-form them. The script re-forms based on matching primary listing full codes and the crosswalk.

### DIFF
--- a/backend/courses/management/commands/form_simple_topics.py
+++ b/backend/courses/management/commands/form_simple_topics.py
@@ -27,7 +27,11 @@ def form_simple_topics():
             list(
                 set(
                     course.topic
-                    for course in primary_listings.filter(full_code=full_code).select_related("topic")
+                    for course in (
+                        primary_listings
+                        .filter(full_code=full_code)
+                        .select_related("topic")
+                    )
                 )
             )
         )
--- a/backend/courses/management/commands/form_simple_topics.py
+++ b/backend/courses/management/commands/form_simple_topics.py
@@ -1,0 +1,79 @@
+from django.core.management.base import BaseCommand
+from django.db import transaction
+from django.db.models import Exists, F, OuterRef
+
+from alert.management.commands.recomputestats import garbage_collect_topics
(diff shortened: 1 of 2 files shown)


backend/courses/management/commands/form_simple_topics.py
```diff
@@ -0,0 +1,81 @@
...
+from alert.management.commands.recomputestats import garbage_collect_topics
+from courses.management.commands.load_crosswalk import get_crosswalk_s3
+from courses.models import Course, Topic
+
+
+def form_simple_topics():
+    Course.objects.all().update(topic=None)
+    garbage_collect_topics()
+
+    print("Cleared topics.")
```
shiva-menta, line 14: Should we have these print statements only run `if verbose`?

backend/courses/management/commands/form_simple_topics.py
```diff
@@ -0,0 +1,81 @@
...
+    # create topics for each primary listing
+    primary_listings = Course.objects.filter(primary_listing_id=F("id"))
+    for primary_listing in primary_listings:
+        topic = Topic.objects.create(most_recent=primary_listing)
+        primary_listing.listing_set.all().update(topic=topic)
+
+    print("Created new topics (one per course).")
+
+    # merge topics based on full_code of primary_listing
+    for full_code in primary_listings.values_list("full_code", flat=True):
```
shiva-menta, line 25: Would it be easier to use `collections.defaultdict(set)` based on full_code, then iterate through the dict and merge topics on each individual set?

Could be less iteration if I'm understanding correctly. but don't think it's that big of a difference.

AaDalal: I think it would end up being about the same work

backend/courses/management/commands/form_simple_topics.py
AaDalal: This is a good point but I broke my dev env and I'm afraid of making any changes without testing locally :( will add it to my to do list though.

PR #512, opened 2023-09-21
Change topics to soft state, derive from parent_course graph

mureytasroc:
This pull request re-architects the Topics data model based on @AaDalal's idea to use `parent_course` relationships between courses to derive topics (thus topics are just soft state derived from the `parent_course` graph).

It also renames the important `recomputestats` script to `recompute_soft_state` to more accurately describe the job of this script. It basically recomputes any major state that is derived from other sources of truth in our database. If we didn't care about efficiency of DB queries, `recompute_soft_state` would not have to exist (however, precomputing some soft state is necessary for tractable query performance).

Finally, it changes PCR to optionally take a semester path parameter to indicate you want review data for a superseded course (a course offered in a past semester that has more recently been replaced by another course with the same full code but different topic).


backend/PennCourses/settings/base.py
```diff
@@ -210,6 +210,10 @@
 XWALK_S3_BUCKET = "penn.courses"
 XWALK_SRC = "xwalk_csre_number.txt"
 
+# The first semester that used Banner/NGSS for course data
+# Note that the above crosswalk connects pre and post NGSS courses
+FIRST_BANNER_SEM = "2022B"
```
AaDalal, line 215: might be good to specify this is when course codes changed for posterity

backend/courses/management/commands/recompute_topics.py
```diff
@@ -0,0 +1,95 @@
...
+            for course in tqdm(courses):
+                parent = course.parent_course
+                if not parent:
+                    course.topic = Topic(most_recent=course)
+                    course.topic.save()
+                    continue
+                if (
+                    parent.full_code == course.full_code
+                    and parent.title == course.title
+                    or len(parent.children) == 1
```
AaDalal, line 49: Should this be parent.children.count() instead? or does this form work?

backend/courses/management/commands/recompute_topics.py
```diff
@@ -0,0 +1,95 @@
...
+                ):
+                    # If a parent has multiple children and none are an exact match, we consider
+                    # all the child courses as "branched from" the old. But if one child is an exact
+                    # match, we inherit the parent's topic and the rest of the children
+                    # are considered "branched from".
+                    if course.topic_id != parent.topic_id:
+                        course.topic = course.parent_course.topic
```
AaDalal, line 56: why not say `parent.topic` instead of `course.parent_course.topic`?

backend/courses/views.py
```diff
@@ -246,13 +263,22 @@ def get_queryset(self):
                 ),
             )
         )
+        check_semester = self.request.GET.get("check_semester")
```
AaDalal, line 266: 1. What's the use case for check_semester? 
2. if there is a strong one, is there anything we could rename to that better captures what it does? (`check_offered_in` or something similar?)

Also, nit: can't we use `self.request.query_params.get(...)`?

backend/review/documentation.py
```diff
@@ -132,6 +132,10 @@
                         "type": "string",
                         "description": "The dash-joined department and most-recent code of this course, e.g. `CIS-1200`.",  # noqa E501
                     },
+                    "superseded_last_offered": {
```
AaDalal, line 135: This field is kind of confusing. Is there anything to do about the naming here? I'm coming up blank right now.

backend/review/views.py
```diff
@@ -82,26 +95,12 @@ def extra_metrics_section_filters_pcr(current_semester=None):
...
     if current_semester is None:
         current_semester = get_current_semester()
-    return (
-        extra_metrics_section_filters
-        & Q(course__primary_listing_id=F("course_id"))
-        & Q(course__semester__lt=current_semester)
-        & ~Q(status="X")
-    )
-
+    return extra_metrics_section_filters & Q(course__semester__lt=current_semester) & ~Q(status="X")
```
AaDalal, line 98: Why remove this filter? Couldnt' that lead to crosslisted courses being double counted?

backend/review/views.py
```diff
@@ -125,39 +124,95 @@ def extra_metrics_section_filters_pcr(current_semester=None):
...
 @permission_classes([IsAuthenticated])
-def course_reviews(request, course_code):
+def course_reviews(request, course_code, semester=None):
     """
     Get all reviews for the topic of a given course and other relevant information.
     Different aggregation views are provided, such as reviews spanning all semesters,
     only the most recent semester, and instructor-specific views.
     """
     try:
+        semester = request.GET.get("semester")
```
AaDalal, line 151: Nit about request.query_params.get()

backend/review/views.py
```diff
@@ -125,39 +124,95 @@ def extra_metrics_section_filters_pcr(current_semester=None):
...
         course = (
-            Course.objects.filter(course_filters_pcr_allow_xlist, full_code=course_code)
+            Course.objects.filter(
+                course_filters_pcr,
+                **(
+                    {"topic__courses__full_code": course_code, "topic__courses__semester": semester}
+                    if semester
+                    else {"full_code": course_code}
+                ),
```
AaDalal, line 159: I'm not understanding why we don't use the topic when the semester is not specified. Would you mind explaining?

backend/review/views.py
```diff
@@ -125,39 +124,95 @@ def extra_metrics_section_filters_pcr(current_semester=None):
...
+    superseded = (
+        Course.objects.filter(
+            course_filters_pcr,
+            full_code=course_code,
+        )
+        .aggregate(max_semester=Max("semester"))
+        .get("max_semester")
+        > course.semester
+        if semester
+        else False
```
AaDalal, line 192: Doesn't this code imply you are superseded if there is another course within the same lineage? But that's not how I understood the definition of superseded from earlier...

Would you mind elaborating on why this depends on `semester` and why we compare to the `max_semester`?

backend/review/views.py
```diff
@@ -582,12 +666,10 @@ def instructor_for_course_reviews(request, course_code, instructor_id):
     check_instructor_id(instructor_id)
     instructor = get_object_or_404(Instructor, id=instructor_id)
 
-    topic = course.topic
-    course = course.topic.most_recent
-
     reviews = review_averages(
         Review.objects.filter(
-            review_filters_pcr, section__course__topic=topic, instructor_id=instructor_id
+            section__course__topic_id=course.topic_id,
```
AaDalal, line 671: Are these changes related to the other ones or are these just general optimizations?

backend/review/views.py
```diff
@@ -666,17 +748,26 @@ def autocomplete(request):
...
+        .values("full_code", "title", "max_semester")
         .distinct()
     )
     course_set = sorted(
         [
             {
                 "title": course["full_code"],
                 "desc": [course["title"]],
-                "url": f"/course/{course['full_code']}",
+                "url": f"/course/{course['full_code']}/{course['max_semester']}",
```
AaDalal, line 770: Why is the semester appended to the end of the URL here?

AaDalal (commented):
Overall, I think it would be good to talk through some of the code along the following axes:
1. How are we making/creating the soft state? How are we creating the hard state?
2. What code is related to the soft/hard state vs what is not?

I think I might be able to get a better understanding of this PR if we discuss those things; it would also probably be helpful to discovering any edge cases.

backend/alert/management/commands/alertstats.py
```diff
@@ -15,7 +15,7 @@
 class Command(BaseCommand):
     help = (
         "Get statistics on PCA, and optionally send to Slack (for analytics use only; do "
-        "not confuse this script with the recomputestats command, which actually updates "
+        "not confuse this script with the recompute_soft_state command, which actually updates "
```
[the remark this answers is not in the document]
mureytasroc: It can also be run manually though - any many internally critical cronjobs like iscimport don't have that convention so I would say we should either do it for all such scripts or none

backend/alert/tasks.py
```diff
@@ -68,7 +68,7 @@ def send_course_alerts(course_code, course_status, semester=None, sent_by=""):
 
 @shared_task(name="pca.tasks.recompute_percent_open")
 def recompute_percent_open_async(semester):
-    recompute_percent_open(semesters=[semester], semesters_precomputed=True)
+    recompute_percent_open(semesters=[semester])
```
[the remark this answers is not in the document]
mureytasroc: Yeah I removed that though so now everything assumes the semesters are precomputed

backend/courses/management/commands/fill_parent_courses.py
```diff
@@ -0,0 +1,81 @@
...
+        print(f"Attempting to fill empty `parent_course` fields for semesters {semesters}.")
+    filled_links = 0
+    for i, semester in enumerate(sorted(semesters)):
+        if verbose:
+            print(f"Processing semester {semester} ({i+1}/{len(semesters)})...")
+        with transaction.atomic():
+            courses = list(Course.objects.filter(semester=semester, parent_course__isnull=True))
+            full_codes = {course.full_code for course in courses}
+            parent_candidates = (
+                Course.objects.filter(full_code__in=full_codes, semester__lt=semester)
```
mureytasroc, line 28: Should we allow for small variations in full code (e.g. prepending or appending 0)?

backend/courses/management/commands/recompute_soft_state.py
```diff
@@ -581,19 +473,26 @@ def add_arguments(self, parser):
...
             default=None,
         )
 
     def handle(self, *args, **kwargs):
         root_logger = logging.getLogger("")
         root_logger.setLevel(logging.DEBUG)
 
-        recompute_stats(semesters=kwargs["semesters"], verbose=True)
+        print(
+            "This script is atomic per step, per-semester. "
```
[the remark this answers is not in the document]
mureytasroc: True maybe we should just not print any atomicity guarantees bc this is soft state anyway

backend/courses/management/commands/recompute_topics.py
```diff
@@ -0,0 +1,95 @@
...
+
+from courses.management.commands.recompute_soft_state import garbage_collect_topics
+from courses.models import Course, Topic
+from courses.util import all_semesters
+
+
+def recompute_topics(min_semester: str = None, verbose=False):
+    """
+    Course topics are directly derived from the `Course.parent_course` graph.
+        - Any orphan child gets its own topic.
```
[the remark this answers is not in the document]
mureytasroc: Orphan meaning it doesn't have a parent... I'll make this clearer

backend/courses/management/commands/recompute_topics.py
```diff
@@ -0,0 +1,95 @@
...
+        - Any single child of a parent inherits the parent's topic.
+        - Any child with siblings that exactly matches their parent in full_code
+          and title inherits the parent's topic.
+        - Any child with siblings that does not exactly match their parent
```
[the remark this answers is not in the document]
mureytasroc: I think any edge cases where someone changes full_code and title in the same semester as adding multiple new children of a parent course, can be addressed manually. Generally in these cases it seems like the children are deviating too much from the original parent to consider them as being in the same topic.

backend/courses/management/commands/recompute_topics.py
mureytasroc: I think this works because children are prefetched (and even if they weren't, I think this would still trigger a query like count). I'll change this to count though.

backend/courses/management/commands/fill_parent_courses.py
AaDalal: I think this makes sense since this is generally a heuristic used by Wharton, CIS and others

backend/courses/management/commands/fill_parent_courses.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+    if verbose:
+        print(f"Attempting to fill empty `parent_course` fields for semesters {semesters}.")
+    filled_links = 0
+    for i, semester in enumerate(sorted(semesters)):
+        if verbose:
+            print(f"Processing semester {semester} ({i+1}/{len(semesters)})...")
+        with transaction.atomic():
+            courses = list(Course.objects.filter(semester=semester, parent_course__isnull=True))
+            full_codes = {course.full_code for course in courses}
```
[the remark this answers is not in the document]
mureytasroc: I think that would be a case we have to manually address. I don't see any way to automatically distinguish that from the case where the original code is just not offered in those intervening years.

backend/courses/management/commands/fill_parent_courses.py
AaDalal: How are we planning on using the crosswalk?

backend/courses/migrations/0034_auto_20211114_0032.py
```diff
@@ -15,15 +14,15 @@ class Migration(migrations.Migration):
...
+                help_text="\nThe number of distinct activities belonging to this course (precomputed for efficiency).\nMaintained by the registrar import / recompute_soft_state script.\n",
             ),
         ),
         migrations.AddField(
             model_name="section",
             name="num_meetings",
             field=models.IntegerField(
                 default=0,
-                help_text="\nThe number of meetings belonging to this section (precomputed for efficiency).\nMaintained by the registrar import / recomputestats script.\n",
+                help_text="\nThe number of meetings belonging to this section (precomputed for efficiency).\nMaintained by the registrar import / recompute_soft_state script.\n",
```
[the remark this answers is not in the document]
mureytasroc: It would create new migrations

backend/courses/views.py
mureytasroc: This is so that PCR URLs can now contain the semester of the course (since course codes may change) - if someone specifies a semester in the URL, they are saying "I want all review data for the topic with this course code in this semester", which is guaranteed to be unique. Then, we only want to fetch current data for the most recent course code in the topic, but only if it is was offered in that semester... otherwise the currently offered course is not in the same topic.

backend/courses/migrations/0034_auto_20211114_0032.py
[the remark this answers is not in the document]
mureytasroc: Sry fixed

backend/courses/management/commands/recompute_topics.py
```diff
@@ -0,0 +1,104 @@
...
+            courses_to_update = []  # to update .topic
+            topics_to_update = []  # to update .most_recent
+            for course in tqdm(courses):
+                parent = course.parent_course
+                if not parent:
+                    course.topic = Topic(most_recent=course)
+                    course.topic.save()
+                    continue
+                if (
+                    parent.full_code == course.full_code
```
AaDalal, line 56: This is speculative, but do you know how the crosswalk handles crosslistings? Seems like if it includes mappings from the same 3 digit code to multiple crosslistings we could end up with the topic explosion we were hoping to avoid...

frontend/review/src/pages/ReviewPage.js
```diff
@@ -118,7 +121,7 @@ export class ReviewPage extends Component {
           } else {
             this.setState({ data });
             if (type === "course") {
-              apiLive(data.code)
+              apiLive(data.code, semester)
                 .then(result => this.setState({ liveData: result }))
                 .catch(() => undefined);
```
AaDalal, line 126: Why is this catch empty: shouldn't it put up some warnings about the course not being offered in the given semester?

backend/courses/views.py
```diff
@@ -246,13 +263,22 @@ def get_queryset(self):
                 ),
             )
         )
+        check_offered_in = self.request.query_params.get("check_offered_in")
```
AaDalal, line 266: Why use check offered in instead of just changing the semester slug param?

AaDalal (changes requested):
Went through this with a fine tooth comb. 
1. Still don't love the API for check_offered_in/last_offered_sem_if_superseded, but I don't have a complete solution there. 
2. Given that, I think this looks pretty close to being ready to merge.
3. Remaining tasks seem mainly to test this (I am dumping the prod DB right now so we can try running on that) and fix some of these nits.

backend/review/views.py
mureytasroc: The idea is now crosslisted courses will be in separate topics so we don't want to count those out

backend/review/views.py
mureytasroc: If the semester isn't specified, we just take the most recent course matching the full code. If the semester is specified, we find the most recent course in the same topic as a course matching the given full code in the given semester.

backend/review/views.py
mureytasroc: `course` is always the most recent course in the topic. So superceded just determines if that full code has been offered in a strictly later semester, indicating the course has been superceded by a different topic

mureytasroc: if `semester` isn't specified, we always take the most recent course matching the full code no matter what so that check will always evaluate to False

AaDalal: Got it makes sense.

backend/courses/management/commands/recompute_parent_courses.py
```diff
@@ -0,0 +1,88 @@
...
+            if verbose:
+                print(f"Recomputed {len(courses)} `parent_course` links.")
+            filled_links += len(courses)
+    if verbose:
+        print(
+            f"Finished recomputing {filled_links} `parent_course` fields for semesters {semesters}."
+        )
+
+    try:
+        load_crosswalk(print_missing=False, verbose=verbose)
```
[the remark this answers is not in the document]
mureytasroc: Hmm yes BUT I don't think we can use the crosswalk to delete existing parent links because the crosswalk doesn't claim to have a comprehensive mapping of parent to child courses...

I believe the soft state topic script should handle multiple children fairly well. We could theoretically have some boolean flag denoting whether a link came from the crosswalk / manual input vs our automatic heuristic (rn just same full code)... that might be a good way to prevent this automatic script from overwriting manually confirmed links.

backend/courses/management/commands/recompute_topics.py
mureytasroc: Wow, GREAT catch. Just checked the crosswalk and it has the confusing behavior of mapping a single old code to all the new crosslistings, rather than mapping crosslistings separately like `OIDD-291 -> OIDD-2910` and `MGMT-291 -> MGMT-2910`. Modified the load crosswalk script to split up crosslistings appropriately.

backend/courses/views.py
mureytasroc: Changing the semester slug param will change the returned data, whereas check offered in is just a guard that will return 404 if the course wasn't offered in a certain semester, but won't change the returned data

frontend/review/src/pages/ReviewPage.js
mureytasroc: I think the user gets that feedback from not seeing the live data UI in PCR (stars, sections, status, etc). Don't think an explicit modal is necessary

backend/review/views.py
```diff
@@ -82,24 +101,14 @@ def extra_metrics_section_filters_pcr(current_semester=None):
...
 
-section_filters_pcr = Q(course__primary_listing_id=F("course_id")) & (
-    Q(has_reviews=True)
-    | ((~Q(course__title="") | ~Q(course__description="")) & ~Q(activity="REC") & ~Q(status="X"))
+course_filters_pcr = ~Q(title="") | ~Q(description="") | Q(sections__has_reviews=True)
+section_filters_pcr = Q(has_reviews=True) | (
+    (~Q(course__title="") | ~Q(course__description="")) & ~Q(activity="REC") & ~Q(status="X")
 )
-
 review_filters_pcr = Q(section__course__primary_listing_id=F("section__course_id"))
```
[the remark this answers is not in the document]
mureytasroc: Yeah this should have been deleted, not sure how it got re-added

backend/review/views.py
```diff
@@ -125,39 +134,95 @@ def extra_metrics_section_filters_pcr(current_semester=None):
...
-            .select_related(
-                "topic",
-                "topic__most_recent",
-                "topic__branched_from",
-                "topic__branched_from__most_recent",
+            .annotate(
+                aliases=Subquery(
+                    Course.objects.filter(primary_listing=OuterRef("primary_listing"))
+                    .exclude(id=OuterRef("id"))
+                    .values_list("full_code", flat=True)
```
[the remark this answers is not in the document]
mureytasroc: Shoot, no you can't. Fixed in next commit

backend/review/views.py
mureytasroc: Just general optimizations

backend/review/views.py
```diff
@@ -666,17 +758,32 @@ def autocomplete(request):
...
+                .values("max_semester")
+            )
+        )
+        .filter(semester=F("max_semester"))
+        .values("full_code", "title", "topic_id", "max_semester")
+        .distinct("full_code", "topic_id")
+    )
+    code_title_counter = Counter(courses.values("full_code", "title"))
+    semester_prefix = (
+        lambda course: f"({course.semester})  "
```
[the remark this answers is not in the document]
mureytasroc: Good catch

backend/review/views.py
mureytasroc: I think it's fine to fully qualify the URL in links clicked by search results - then anyone who copies the link from the URL will have a permalink to that specific course for that topic, rather than future superceding courses. The returned page will be the same as long as the course hasn't been superceded.

backend/courses/management/commands/fill_parent_courses.py
mureytasroc: The NGSS crosswalk? That's loaded at the end of this script.

For now let's just stay on the extra careful side and only link 3-4 digit codes if it's included in the crosswalk or we manually add a `parent_course` link. We really don't want to continue to be plagued by the topic blowup problem.

backend/alert/management/commands/compute_head_registrations.py
```diff
@@ -1,40 +0,0 @@
-import gc
-from textwrap import dedent
-
-from django.core.management.base import BaseCommand
-from tqdm import tqdm
-
-from alert.models import Registration
-
-
-class Command(BaseCommand):
```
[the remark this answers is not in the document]
mureytasroc: Yeah we only needed to run this script once when the auto-resubscribe migration was originally made a long time ago - no need for it now.

PR #514, opened 2023-09-23
Correctly apply phone number validation on PCA backend

mureytasroc:
The PCA backend didn't used to actually validate phone numbers, as `phonenumbers.parse` doesn't validate. Changed the `UserProfile` validator to properly validate on serializer create (and `object.full_clean()` - NOT `object.save()` btw) and added some tests.

### DIFF
--- a/backend/courses/models.py
+++ b/backend/courses/models.py
@@ -1369,7 +1369,6 @@ class UserProfile(models.Model):
             return
         try:
             parsed_number = phonenumbers.parse(value, "US")
-            print(parsed_number)
             if not phonenumbers.is_valid_number(parsed_number):
                 raise ValueError(f"Invalid phone number '{value}'.")
         except (phonenumbers.phonenumberutil.NumberParseException, ValueError) as e:
--- a/backend/courses/models.py
+++ b/backend/courses/models.py
@@ -1359,14 +1359,21 @@ class UserProfile(models.Model):
         """
         Validator to check that a phone number is in the proper form. The number must be in a
         form which is parseable by the
-        [phonenumbers library](https://pypi.org/project/phonenumbers/).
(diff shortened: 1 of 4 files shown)


backend/courses/models.py
```diff
@@ -1396,16 +1402,14 @@ def save(self, *args, **kwargs):
...
-        if self.phone is not None:
-            try:
-                phone_number = phonenumbers.parse(self.phone, "US")
-                self.phone = phonenumbers.format_number(
-                    phone_number, phonenumbers.PhoneNumberFormat.E164
-                )
-            except phonenumbers.phonenumberutil.NumberParseException:
-                raise ValidationError("Invalid phone number (this should have been caught already)")
+        if self.phone:
+            # self.phone validated by `validate_phone`
```
AaDalal, line 1408: Nit but I think this should be 'should be validated' since technically it's not guaranteed the validator is called (https://docs.djangoproject.com/en/4.2/ref/models/instances/#django.db.models.Model.clean_fields)

backend/tests/courses/test_api.py
```diff
@@ -716,14 +716,14 @@ def test_patch(self):
         self.assertEqual(response.data["first_name"], "new_name")
         response = self.client.patch(
             "/accounts/me/",
-            json.dumps({"profile": {"phone": "3131234567"}}),
+            json.dumps({"profile": {"phone": "19178286431"}}),
```
AaDalal, line 719: What's the reason for changing these numbers?

AaDalal (approved):
Lgtm

Left a couple of questions + very tiny nits

backend/tests/courses/test_api.py
mureytasroc: They were invalid for some reason

PR #515, opened 2023-09-27
Update iscimport script to handle banner semesters (e.g., '202310')

AaDalal:


AaDalal:
@mureytasroc side note, but I think we should re-write this script at some point because it's a little hard to read/maintain RN.

mureytasroc (approved):
Looks good

PR #517, opened 2023-10-14
feat(temporary banners)

esinx:
Screenshot 2023-10-14 at 5 55 30 PM
Screenshot 2023-10-14 at 5 55 10 PM
Screenshot 2023-10-14 at 5 54 00 PM

### DIFF
--- a/frontend/review/src/components/Announcement.js
+++ b/frontend/review/src/components/Announcement.js
@@ -1,0 +1,197 @@
+import React, { useEffect, useRef, useState } from "react"
+
+const issueIcon =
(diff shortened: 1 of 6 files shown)


frontend/review/src/components/Announcement.js
```diff
@@ -0,0 +1,197 @@
+import React, { useEffect, useRef, useState } from "react"
+
```
andyjiang3, line 2: Does `pcx-shared-components` not work on pcr?

andyjiang3 (approved):
lgtm for temp banners! we currently have separate banners for each apps (ex. [pca](https://github.com/pennlabs/penn-courses/blob/1d0fc3ec697914a7aa5c63f53249cf40469d67ad/frontend/alert/pages/index.tsx#L214)) that would need to be deleted after migration to banner for all web products (can do this in a diff pr).

andyjiang3:
Also looks like you need to `yarn clean` to fix the linting issue

frontend/review/src/components/Announcement.js
esinx: I assumed that from the [README](https://github.com/pennlabs/penn-courses/tree/master/frontend). Let me trying moving that into shared components instead, but I'm not sure how using TypeScript in PCR will play out here!

PR #519, opened 2023-10-23
Unset primary schedule feature, fix autodocs

mureytasroc:
This PR allows users to unset their primary schedule, fixes some bugs, and improves testing.

PR #520, opened 2023-10-25
Course audit feature

JeffersonDing:
There are some changes to the docker container environment. Please restart the development environment if things are not working.


backend/courses/views.py
```diff
@@ -524,3 +525,46 @@ def delete(self, request):
...
+class AuditView(generics.ListAPIView):
+    """
+    This route allows you to list courses that are currently happening with filters to
+    exclude departments, buildings etc. The route will return a list of courses in
+    decreasing course_quality. Without GET parameters, this route will return all
+    currently running courses.
+    """
+
+    def get(self, request, semester):
+        timestamp = request.GET.get("timestamp")
```
AaDalal, line 539: Nit: DRF (Django rest framework) provides the request.query_params.get(...) which is more readable

andyjiang3:
Note that this contains some commits from another open [PR](https://github.com/pennlabs/penn-courses/pull/519) that shouldn't be merged yet, prob should remove if we're planning on merging this before that PR

andyjiang3:
Also should prob fix the pre-commit hooks so that it does the linting - seems like we all have different formatters and it makes review harder. @AaDalal (I can look into this but would have to be sometimes late next week)

PR #535, opened 2023-10-27
🔧 Created Makefile and updated README

JeffersonDing:


backend/README.md
```diff
@@ -63,25 +67,34 @@ NOTE: when using `pipenv`, environment variables are only refreshed when you exi
...
+- Rename the sql dump you downloaded to `pcx_test.sql` and place it in the `backend` directory. Then run `make load` to load the data into your local database. This might take a while.
 - For accessing the Django admin site, the admin username is `admin` and the password is `admin` if you use this test db.
 
+[//]: # (- Clear the existing contents of your local database with `psql template1 -c 'drop database postgres;' -h localhost -U penn-courses` &#40;the password is `postgres`&#41;)
+
+[//]: # (- Create a new database with `psql template1 -c 'create database postgres with owner "penn-courses";' -h localhost -U penn-courses` &#40;same password&#41;.)
+
+[//]: # (   )
+
+[//]: # (- Finally, run `psql -h localhost -d postgres -U penn-courses -f pcx_test.sql` &#40;replacing `pcx_test.sql` with the full path to that file on your computer&#41; to load)
```
AaDalal, line 93: Why is it formatted like this ?

backend/Makefile
```diff
@@ -0,0 +1,52 @@
+# Commands
+runserver:
+	PIPENV_VERBOSITY=-1
+	pipenv run python manage.py runserver 0.0.0.0:8000
```
AaDalal, line 4: Nit, but why not just use 127.0.0.1?

backend/README.md
```diff
@@ -9,6 +9,10 @@ Welcome to the Penn Courses Backend (PCX)!
 
 ### Running the Backend with Docker-Compose
```
AaDalal, line 10: I think we should move the local dev environment thing to the top, and move down `Running the backend with Docker-Compose`.

backend/README.md
```diff
@@ -153,50 +166,28 @@ If you don't want to use docker alone, you can also set up and run the dev envir
...
       > - (WSL) `[sudo] service docker start`
 
-4. Setting up your Penn Courses development environment
-
-   1. `pipenv install --dev`
-   2. `pipenv shell`
-   3. `python manage.py migrate`
-
-5. Loading test data (if you are a member of Penn Labs). If you are not a member of Penn Labs, you can skip this section and load in course data from the registrar, as explained below.
+4. Migrate Django models
```
AaDalal, line 177: Hmm, I think using make for simple things like migrate/makemigrations could end up being a leaky abstraction as we will not provide make commands for all the commands they might run (e.g., management commands or manage.py shell). 

IMO, I think we should add make commands for the installation, but otherwise ask people to use `python manage.py xxx`.

backend/Makefile
```diff
@@ -0,0 +1,52 @@
...
+# Make migrations (when models are changed)
+makemigrations:
+	python manage.py makemigrations
+
+# Migrate (when migrations are created)
+migrate:
+	python manage.py migrate
+
+# Lint
+lint:
```
AaDalal, line 25: make lint is useful

backend/Makefile
```diff
@@ -0,0 +1,52 @@
...
+# Launch docker database & cache
+launchdb:
+	docker compose up -d
+
+# Make migrations (when models are changed)
+makemigrations:
+	python manage.py makemigrations
+
+# Migrate (when migrations are created)
+migrate:
```
AaDalal, line 21: I don't think we should abstract away the python manage.py xxx commands (see my comment on the readme)

AaDalal (changes requested):

PR #539, opened 2023-11-02
Schedule Solver Feature

el-agua:
Added endpoint and tests. More detailed commits are on the automatic-scheduler branch, but the linter kinda messed it up.

### DIFF
--- a/backend/courses/util.py
+++ b/backend/courses/util.py
@@ -1,5 +1,5 @@
-import json
 import itertools
+import json
 import logging
 import os
 import random
--- a/backend/courses/util.py
+++ b/backend/courses/util.py
@@ -686,6 +686,7 @@ def find_possible_schedules(
     """
 
     day_to_num = {"M": 0, "T": 1, "W": 2, "R": 3, "F": 4}
+    whitelist = ["CIS-7000"]
 
     class Scheduler:
         def __init__(self):
@@ -744,10 +745,14 @@ def find_possible_schedules(
             for interval in optimal_schedule:
                 class_name = interval[2]
                 sections = [i for i in optimal_schedule if i[2] == class_name]
-                section = random.choice(sections)
-                if class_name not in selected_classes:
+                if class_name not in whitelist:
+                    section = random.choice(sections)
+                    if class_name not in selected_classes:
+                        selected_classes.add(class_name)
+                        final_schedule.append(section)
+                else:
                     selected_classes.add(class_name)
-                    final_schedule.append(section)
+                    final_schedule += sections
             return final_schedule
 
     def find_sections(courses):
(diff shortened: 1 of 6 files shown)


backend/courses/util.py
```diff
@@ -672,3 +674,352 @@ def get_semesters(semesters=None, verbose=False):
...
+            courses.append(random.choice(hash[key]))
+        return courses
+
+    def credit_count(sections):
+        acc = 0
+        for section in sections:
+            acc += section[1]
+        return acc
+
+    def check_if_schedule_possible(schedule, courses, unwanted_intervals):
```
shiva-menta, line 867: Is the logic too complicated to incorporate this directly until whatever interval solver problem you have? Trying to determine the real need for a separate function.

backend/plan/urls.py
```diff
@@ -15,6 +16,7 @@
 router.register(r"primary-schedules", PrimaryScheduleViewSet, basename="primary-schedules")
 
 urlpatterns = [
+    path("scheduler/courses/", AutomaticCourseScheduler.as_view(), name="automatic-scheduler"),
```
shiva-menta, line 19: Probably don't need the `/courses/` unless you think there is some collision that might come up.

shiva-menta (commented):
Looks solid, just left a few points of feedback!

backend/plan/urls.py
el-agua: Fixed.

backend/courses/util.py
el-agua: There might be a better way to do this, but the use of it is to resolve conflicts after merging all the days together. It also makes adding recitations a little cleaner, as its kinda like "induction": if schedule works, does schedule + recs work?

backend/courses/util.py
```diff
@@ -672,3 +674,352 @@ def get_semesters(semesters=None, verbose=False):
...
+
+    def remove_duplicates(given_l):
+        """
+        Removes duplicates from a list.
+        Example:
+        remove_duplicates([1,1,1,3,5,7,7]) = [1,3,5,7]
+        """
+        newl = []
+        [newl.append(x) for x in given_l if x not in newl]
+        return newl
```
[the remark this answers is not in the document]
el-agua: String lists aren't hashable

backend/courses/util.py
```diff
@@ -672,3 +674,352 @@ def get_semesters(semesters=None, verbose=False):
...
+            return True
+
+        def find_optimal_schedule(self, unwanted_intervals=[]):
+            # Sort intervals by end time
+            sorted_intervals = sorted(self.intervals, key=lambda x: x[1])
+            # Remove unwanted intervals
+            not_ok = []
+            for unwanted_interval in unwanted_intervals:
+                [
+                    not_ok.append(interval[3])
```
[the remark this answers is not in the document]
el-agua: Fixed with whitelist

PR #560, opened 2023-12-23
Add attributes filter to PCP

AaDalal:
This PR adds support for course attributes to PCP, including an attribute filter and adding attributes to course details. Attirbutes includes Wharton, SEAS, SAS, and Nursing, as well as additional ones for GSE, LPS, Grade Mode etc.

TODO:
- [ ] change color of checkboxes and radio buttons
- [ ] fix issue of page overflowing when there are many attributes available
- [ ] fix attributes selection dropdown auto closing after selection


frontend/plan/types.ts
```diff
@@ -1,8 +1,15 @@
...
     SEAS = "SEAS",
     WHARTON = "WH",
     COLLEGE = "SAS",
     NURSING = "NURS",
+    LPS = "LPS",
+    DESIGN = "DSGN",
+    GSE = "GSE",
+    LAW = "LAW",
+    MEDICINE = "MED",
+    GRADE_MODE = "MODE", // this isn't really a school, but it's used for course attributes
```
esinx, line 11: Is this something we could split using union types? or is it better left as a one-off? Would love to hear more background behind this decision!

frontend/plan/types.ts
```diff
@@ -177,7 +182,7 @@ export interface Friendship {
 export interface FilterData {
     searchString: string;
     searchType: string;
-    selectedReq: { [K in string]: boolean };
+    selectedAttrs: { [K in string]: boolean };
```
esinx, line 185: Can we make this type more strict by defining what attributes are available?

esinx (commented):
Minor suggestions & type-related questions!

frontend/plan/types.ts
AaDalal: 2 reasons
1. blindly following the conventions of the rest of this file (it was probably written before union types?)
2. easier to guarantee exhaustive matching (?)

I'm not a typescript expert, so I'll defer to you

frontend/plan/types.ts
AaDalal: Attributes are dynamic (e.g., we fetch them from the database) so likely not

PR #567, opened 2024-01-14
Replace pipenv with poetry

Clue88:


.github/workflows/shared-build-deploy.yaml
```diff
@@ -9,13 +9,14 @@ on: push
 jobs:
   backend-check:
     name: "Backend Check"
-    uses: pennlabs/shared-actions/.github/workflows/django.yaml@8785a7d7b9158d8d5705a0202f5695db2c0beb97
+    uses: pennlabs/shared-actions/.github/workflows/django.yaml@7b67c73ced6fa70043451f6816c706a003642828
```
AaDalal, line 12: having to pin like this sucks because its pretty opaque :(

PR #576, opened 2024-02-04
Updated caching strategy

AaDalal:
Introduces a caching strategy to reduce database load.
- Converts caching on `course_reviews` function to be manual instead of using Django decorator. This allows us to be more efficient with cache usage. We now store mappings from `course -> topic` and `topic -> review` in the cache instead of `course -> review`.
- Adds `precompute_pcr_views` management command to precompute review summaries whenever topics are rearranged or new review data comes in. Review summaries are stored in new `CachedReviewResponse` objects.
- The `precompute_pcr_views` management command is called during `iscimport` and `registrarimport` to ensure review data is as up-to-date as possible.


backend/review/management/commands/precompute_pcr_views.py
```diff
@@ -0,0 +1,101 @@
...
+    responses = CachedReviewResponse.objects.all()
+    topic_set = {response.topic_id: response for response in responses}
+
+    objs_to_insert = []
+    objs_to_update = []
+    cache_deletes = set()
+    has_count = total_count = 0
+
+    # Iterate through all topics.
+    with transaction.atomic():
```
el-agua, line 28: Looks like it works; makes sense

el-agua (approved):
lgtm

More comments would be helpful? might be a more efficient way that doesn't require looping through all topics but looks fine for now.

backend/review/models.py
```diff
@@ -94,6 +94,18 @@ def get_averages(topic_id, instructor_name=None, fields=None):
 ALL_FIELD_SLUGS = [x[2] for x in REVIEW_BIT_LABEL]
 
 
+class CachedReviewResponse(models.Model):
+    """
+    Represents a mapping from temp_topic_id (string-delimited list of sorted courses within a topic)
```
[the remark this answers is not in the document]
shiva-menta: Yep exactly, right now the maximum string length is around 500 chars so this just gives some extra buffer

backend/review/management/commands/iscimport.py
```diff
@@ -277,6 +278,7 @@ def handle(self, *args, **kwargs):
 
         print("Recomputing Section.has_reviews...")
         recompute_has_reviews()
+        precompute_pcr_views(True, True)
```
AaDalal, line 281: nit: maybe make these keyword args

backend/review/management/commands/precompute_pcr_views.py
```diff
@@ -0,0 +1,101 @@
...
+from courses.models import Topic
+from review.models import CachedReviewResponse
+from review.views import manual_course_reviews
+
+
+def precompute_pcr_views(verbose=False, is_new_data=False):
+    if verbose:
+        print("Now precomputing PCR reviews.")
+
+    CachedReviewResponse.objects.all().update(expired=True)
```
AaDalal, line 18: This is not in an atomic block, so what's the potential time in which the cached responses are down totally. Is that acceptable?

backend/review/management/commands/precompute_pcr_views.py
```diff
@@ -0,0 +1,101 @@
...
+                    if is_new_data:
+                        response_obj.response = manual_course_reviews(
+                            topic.most_recent.full_code, topic.most_recent.semester
+                        )
+                        cache_deletes.add(topic_id)
+                    response_obj.expired = False
+                    objs_to_update.append(response_obj)
+                except Http404:
+                    logging.info(
+                        f"Topic returned 404 (topic_id {topic_id}, "
```
AaDalal, line 50: nit: is a 404 an expected behavior? if so, logging.info makes sense; if not logging.error might be better so it looks correct on our sentry

backend/review/management/commands/precompute_pcr_views.py
```diff
@@ -0,0 +1,101 @@
...
+            Topic.objects.all().select_related("most_recent").order_by("most_recent__semester"),
+            disable=not verbose,
+        ):
+            course_id_list, course_code_list = zip(*topic.courses.values_list("id", "full_code"))
+            topic_id = ".".join([str(id) for id in sorted(course_id_list)])
+            total_count += 1
+
+            if topic_id in topic_set:
+                try:
+                    has_count += 1
```
AaDalal, line 39: nit: usually try to minimize the stuff inside the try block to what is absolutely going to throw an error (for debug/code readability reasons)

backend/review/management/commands/precompute_pcr_views.py
```diff
@@ -0,0 +1,101 @@
...
+from review.models import CachedReviewResponse
+from review.views import manual_course_reviews
+
+
+def precompute_pcr_views(verbose=False, is_new_data=False):
+    if verbose:
+        print("Now precomputing PCR reviews.")
+
+    CachedReviewResponse.objects.all().update(expired=True)
+    responses = CachedReviewResponse.objects.all()
```
AaDalal, line 19: nit: I got a little confused bc this is called responses. maybe cached_responses is a better name?

backend/review/management/commands/precompute_pcr_views.py
```diff
@@ -0,0 +1,101 @@
...
+    topic_set = {response.topic_id: response for response in responses}
```
AaDalal, line 20: nit: since this is a dict, and also is a mapping only for cached responses (not for all topics) it might be good to rename it

backend/review/management/commands/precompute_pcr_views.py
AaDalal: NVM i see what you're doing

AaDalal: `expired` CachedReviewResponses can still be returned from the view

backend/review/models.py
```diff
@@ -94,6 +94,18 @@ def get_averages(topic_id, instructor_name=None, fields=None):
...
+    to a JSON object storing summarized course review data (all the data frontend uses to display
+    reviews).
+    """
+
```
AaDalal, line 103: I think topic_id should be unique=True here (theoretically our loading script should enforce this, but I think its good to specify here too?)

backend/review/views.py
```diff
@@ -129,30 +134,59 @@ def extra_metrics_section_filters_pcr(current_semester=None):
...
+    topic_id = cache.get(course_code)
+    if topic_id is None:
+        try:
+            recent_course = most_recent_course_from_code(course_code, request_semester)
+        except Course.DoesNotExist:
+            raise Http404()
+        topic = recent_course.topic
+        course_id_list = list(topic.courses.values_list("id"))
+        topic_id = ".".join([str(id[0]) for id in sorted(course_id_list)])
+        cache.set(course_code, topic_id, MONTH_IN_SECONDS)
```
AaDalal, line 147: why do we set the cache manually?

AaDalal: instead of using the decorator

AaDalal: also we shpuld make sure this cache is cleared every time we reimport and/or that the TTL is short enough.

AaDalal (commented):
Mostly nits / namign things, only 1-2 substantive comments.

Feel free to ignore the nits if you don't have time to fix -- they aren't meant to slow you down!

backend/review/management/commands/precompute_pcr_views.py
shiva-menta: Yeah, but I think there's no reason to not have this in an atomic block either – will change.

backend/review/management/commands/precompute_pcr_views.py
shiva-menta: Iirc there are some topics that have a course with no reviews and that leads to a 404 on our current implementation (can't remember if this is the exact error, but it's something similar). So yeah would say it can be expected behavior on certain courses.

backend/review/views.py
shiva-menta: Cache is set manually because we want some control over setting the course to topic mappings and the topic to response mappings. Using the decorator, since the routes are based on course, would only allow for course to response mapping, which can be a bit space inefficient.

shiva-menta: Yup the relevant cache entries are cleared in the precompute_pcr_reviews script.

PR #582, opened 2024-02-10
Bump react, next and typescript

AaDalal:
This PR upgrades:
1. React to 18.2.0 (prev 16)
2. Next.js to 13.2.1 (prev 9)
3. Typescript to 4.x.x

A few notes
- As a consequence of bumping react, we also bump styled-components. Styled components with react 18 requires using transient props to avoid accidentally leaking props as HTML attributes. Many of the changes in this PR are related to switching styled components to transient props.  
- I ran PCR, PCP and PCA locally with these changes -- all seems good from a UI perspective

Before we merge this pr, we need to merge #583 that bumps node to 18.


esinx:
Works fine on my end, let's just make sure that we have the node version PR in sync

esinx (approved):

frontend/plan/components/search/DayTimeFilter.tsx
```diff
@@ -1,6 +1,6 @@
 import React from "react";
 import styled from "styled-components";
-import { Range } from "rc-slider";
+import Slider from "rc-slider";
```
AaDalal, line 3: Because of rc-slider version bump

AaDalal (commented):
self-review 👍

PR #583, opened 2024-02-10
Bump node to 18

AaDalal:
Bumps node to 18 (previously used 10 for deployment, <=16 in development)

Note: this PR uses node's --openssl-legacy-provider option, which is the same behavior as in Node <= 16. This is insecure, though, so we should fix this in a separate PR.


joyliu-q (approved):
LGTM!

PR #597, opened 2024-02-25
PCX history probabilities for PDP usage

el-agua:
Added historical probabilities management command.

TODO: very basic implementation, looking at bulk_update but unsure if you can run through custom functions?


backend/courses/models.py
```diff
@@ -366,6 +366,36 @@ class Topic(models.Model):
         ),
     )
 
+    historical_probabilities_spring = models.FloatField(
+        default=0,
+        help_text=dedent(
```
AaDalal, line 371: wait why do we need 3 things here. We should just have a single field for the expected semester or unknown (based on some criteria) unless you think there's a strong reason to split it into 3 fields.

el-agua: I think the idea behind this was for the case that even if the expected semester is to capture seasonal classes:
(Ex) If expected semester is Spring 2025, we would have to account for the for the case it is not offered in Fall 2025?

el-agua:
> Mostly just nits. if you've checked that the results are good and this runs in a tractable amount of time, then this looks good to me.
> 
> Can you time how long it takes and lmk? if it's too slow, I think this algo can be rewritten in sql

On my computer it takes around 2 minutes to iterate through the entire test database. For reference, that is about the same time as the current recompute_topic script.

backend/courses/util.py
```diff
@@ -712,3 +712,56 @@ def get_semesters(semesters: str = None) -> list[str]:
             if s not in possible_semesters:
                 raise ValueError(f"Provided semester {s} was not found in the db.")
     return sorted(semesters)
+
+
+def historical_year_probability(current, courses):
+    """
+    current: the current semester represented in the 20XX(A|B|C) format
```
AaDalal, line 719: Super nitpicky: we usually use a specific format for these following the sphinx format

backend/courses/util.py
```diff
@@ -712,3 +712,57 @@ def get_semesters(semesters: str = None) -> list[str]:
...
+    prob_distribution = [0.4, 0.3, 0.15, 0.1, 0.05]
+
+    def normalize_and_round(prob, i):
+        """Modifies the probability distribution to account for the
+        fact that the last course was taken i semesters ago."""
+        truncate = prob_distribution[:i]
+        total = sum(truncate)
+        return list(map(lambda x: round(x / total, 3), truncate))
+
+    def get_semester_and_course_index(semester):
```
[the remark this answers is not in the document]
AaDalal: Can you reply/resolve this?

backend/courses/util.py
```diff
@@ -712,3 +712,57 @@ def get_semesters(semesters: str = None) -> list[str]:
...
+        """Returns an integer representing the semester."""
+        semester_letter = semester[-1]
+        semester_number = 0
+        if semester_letter == "A":
+            semester_number = 1
+        elif semester_letter == "B":
+            semester_number = 2
+        elif semester_letter == "C":
+            semester_number = 3
```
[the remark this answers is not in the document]
AaDalal: Also needs to be replied+resolved

backend/courses/util.py
```diff
@@ -712,3 +712,56 @@ def get_semesters(semesters: str = None) -> list[str]:
...
+    if courses == []:
+        return p
+    else:
+        last_index = int(translate_semester(courses[0].semester)) // 10
+        if last_index > min_index:
+            prob_distribution = normalize_and_round(
+                prob_distribution, ((current_index - last_index) + 9) // 10
+            )
+    for c in courses:
+        """Provides calculation"""
```
AaDalal, line 759: Nit: this doc string is not super useful because the code already indicates this

backend/courses/management/commands/recompute_topics.py
```diff
@@ -151,5 +151,31 @@ def handle(self, *args, **kwargs):
...
+    if verbose:
+        print("Recomputing historical probabilities for all topics...")
+    topics = Topic.objects.all()
+    length = len(topics)
+    # Iterate over each Topic
+    for i, topic in enumerate(topics):
+        if i % 1000 == 0:
+            print(f"Recomputing topics for semesters >={i}/{length}")
+        # Calculate historical_year_probability for the current topic
+        test = topic.courses.order_by("semester").all()
```
AaDalal, line 175: Are you sure you can do the order_by before the all? You might be able to I just don't have personal experience with it so want to make sure it's correct.

AaDalal (changes requested):
I think there were still a few dangling comments from last time + I added some more. You don't have to make changes based on every comment but you should reply when you don't make the changes suggested by the comment, and resolve when you do make the changes suggested

backend/courses/util.py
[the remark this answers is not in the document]
el-agua: Fixed and deleted.

backend/courses/util.py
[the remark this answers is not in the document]
el-agua: Changed to use translate_semester

backend/courses/management/commands/recompute_topics.py
```diff
@@ -151,5 +152,30 @@ def handle(self, *args, **kwargs):
             assert (
                 min_semester in all_semesters()
             ), f"--min-semester={min_semester} is not a valid semester."
-
+        semesters = sorted(
+            [sem for sem in all_semesters() if not min_semester or sem >= min_semester]
+        )
         recompute_topics(min_semester, verbose=True, allow_null_parent_topic=bool(min_semester))
+        if semesters:
```
AaDalal, line 159: In what case do semesters not exist?

backend/courses/management/commands/recompute_topics.py
```diff
@@ -151,5 +152,30 @@ def handle(self, *args, **kwargs):
...
+
+def recompute_historical_semester_probabilities(current_semester, verbose=False):
+    """
+    Recomputes the historical probabilities for all topics.
+    """
+    if verbose:
+        print("Recomputing historical probabilities for all topics...")
+    topics = Topic.objects.all()
+    # Iterate over each Topic
+    for i, topic in tqdm(enumerate(topics)):
```
AaDalal, line 173: should verbose flag apply to tqdm or not? Not sure what pattern we use elsewhere, but you can use disable option

https://stackoverflow.com/questions/37091673/silence-tqdms-output-while-running-tests-or-running-the-code-via-cron

backend/courses/util.py
```diff
@@ -712,3 +712,46 @@ def get_semesters(semesters: str = None) -> list[str]:
...
+        """Modifies the probability distribution to account for the
+        fact that the last course was taken i semesters ago."""
+        truncate = prob_distribution[:i]
+        total = sum(truncate)
+        return list(map(lambda x: round(x / total, 3), truncate))
+
+    current_index = int(translate_semester(current)) // 10
+    min_index = current_index - 60
+    max_index = current_index - 10
+    p = [0, 0, 0]
```
AaDalal, line 739: is it fine that we return an invalid probability distribution in this case?

backend/courses/util.py
```diff
@@ -712,3 +712,46 @@ def get_semesters(semesters: str = None) -> list[str]:
...
+    if courses == []:
+        return p
+    else:
+        last_index = int(translate_semester(courses[0].semester)) // 10
+        if last_index > min_index:
+            prob_distribution = normalize_and_round(
+                prob_distribution, ((current_index - last_index) + 9) // 10
```
AaDalal, line 746: the + 9 here probably warrants a comment bc it's not clear why it's there

backend/courses/util.py
```diff
@@ -712,3 +712,46 @@ def get_semesters(semesters: str = None) -> list[str]:
...
+    for c in courses:
+        index = int(translate_semester(c.semester)) // 10
+        if index < min_index or index > max_index:
+            continue
+        # Diff is the number of years ago the course was taken
+        diff = (current_index - index) // 10 - 1
+        if diff >= len(prob_distribution):
+            diff = len(prob_distribution) - 1
+        p[index % 10 - 1] += prob_distribution[diff]
+    return list(map(lambda x: min(round(x, 2), 1.00), p))
```
AaDalal, line 757: TBH this is pretty complex. Some more comments would do some good (I also wonder if it needs to be this complex? doesn't
```
PROB_DISTRIB = [...] # similar to the one you alr defined
def prob_dist(semesters: list[str], current_semester: str):
    # only include up to 5 years before this one
    semester_probabilities = {
        "a": 0.0,
        "b": 0.0,
        "c": 0.0
    }
    current_year = int(current_semester[:-1])    
    semesters = [semester for semester in semesters if semester < str(current_year) and semester > str(current_year - 5)]
    for historical_semester in semesters:
        historical_year = int(historical_semester[:-1])
        semester_char = historical_semester[-1].upper() # A, B, C
        semester_probabilities[semester_char] = PROB_DIST[current_year - historical_year]
   return normalize(semester_probabilities) # need to renormalize to make semester_probabilitie a proper distribution in case there were skipped years (make sure to handle divide by 0)
```
achieve what we want? or am I missing smth?)

AaDalal (changes requested):
Sorry for the review hell. Will be prompt about re-review on this one

backend/courses/util.py
AaDalal: ok this good is defo not totally correct, but you get the point

backend/courses/util.py
el-agua: The think the idea is that [A,B,C] is not a probability distribution but instead each individual element represents a probability (if the class occurs for the past 5 years for the spring for example, then p[0] = 1)

backend/courses/util.py
el-agua: Added this general paradigm in, tested it and should work.

AaDalal (approved):
LGTM :)

PR #598, opened 2024-02-29
Dev Container

esinx:
To make onboarding faster and trouble-free, we introduce dev containers. In short, onboarding newbies should be as easy as installing Docker *(or [OrbStack](orbstack.dev))*.

After installation, just launch dev container, run `pipenv install --dev` or `yarn` and everything will probably just work™️.

Even `docker compose` works thanks to docker-in-docker. It works surprisingly well with vscode's automatic port forwarding support.

Some caveats are:
- Make sure that your computer can handle multiple docker containers simultaneously
- Use `ssh-add` to "inherit" ssh credentials for source control
- Some initial setup (things like `./manage.py makemigrations`) remains to be done by the user
  - We can actually automate this, not sure if it's a good idea.

Screenshot 2024-02-29 at 4 53 47 PM

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,3 @@
+FROM ubuntu:22.04
+ARG IMAGE_NAME=pennlabs/courses-devcontainer
+RUN apt-get update && apt-get install -y wget curl gcc python3-dev libpq-dev
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,3 +1,3 @@
-FROM ubuntu:22.04
+FROM --platform=linux/amd64 ubuntu:22.04
 ARG IMAGE_NAME=pennlabs/courses-devcontainer
 RUN apt-get update && apt-get install -y wget curl gcc python3-dev libpq-dev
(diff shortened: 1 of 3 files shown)


joyliu-q (approved):
Can you add documentation to the readme on Dev Container setup?
- Devcontainer set-up
- ssh add
- Other caveats mentioned

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM ubuntu:22.04
+ARG IMAGE_NAME=pennlabs/courses-devcontainer
+RUN apt-get update && apt-get install -y wget curl gcc python3-dev libpq-dev
```
joyliu-q, line 3: You should add git here

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM ubuntu:22.04
```
judtinzhang, line 1: After testing some stuff on mobile, also suggest adding `--platform=linux/amd64` to be compatible with our prod machines running on x86 arch

esinx: I'm assuming this will fix dependency resolution? Sounds good to me!

PR #603, opened 2024-03-20
rollback: use kittyhawk 1.1.11 -> 1.1.9 (downgrade to 1.1.9)

esinx:
Redis persistence is causing networking issues between backend and redis, we will roll back to 1.1.9 (doesn't include persistence, previously working build) until we figure out why courses backend cannot talk to redis.

> For future reference, other products(ohq, clubs) are not experiencing this issue despite using kittyhawk@1.1.11


shiva-menta (approved):
lgtm

PR #606, opened 2024-03-22
Removing Redis Stack and Reverting to Base Image

shiva-menta:
Temporarily reverting to base redis since Redis stack was used for PCS, which will be held off until our Fly Migration is complete. This will avoid the issues we've been having with protected mode in `penn-courses-redis`.


esinx (approved):
Fantastic

esinx:
Not that my review is powerful enough to make this PR mergable

PR #607, opened 2024-03-22
Backend Version Upgrades

shiva-menta:
This PR upgrades Django to 5.0.2, Python to 3.11, and makes other minor code changes to adjust for new function signatures (e.g. `construct_search`).

Note that two packages don't officially support Django 5.0: `django-auto-prefetching` and `drf-nested-routers`. Based on the features outlined in Release Notes, this shouldn't pose a problem, but `django-auto-prefetching` doesn't seem to have long-term support, so we should move to a more Django-native solution or another package like `django-auto-prefetch`.


backend/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM pennlabs/django-base:9c4f31bf1af44219d0f9019271a0033a222291c2-3.10.1
+FROM pennlabs/django-base:b269ea1613686b1ac6370154debbb741b012de1a-3.11
```
joyliu-q, line 1: This hash would not work because it tries to install poetry. Let me give you another one

joyliu-q (changes requested):
Rest looks good to me but we need to change the Django base tag because this current one installs poetry and wouldn't work.

joyliu-q (approved):
LGTM, the poetry install was completed but the PATH isn't overridden.

rohangpta:
Moving to `django-auto-prefetch` should be a priority if `django-auto-prefetching` is reaching end of LTS. IIRC courses relies (somewhat heavily) on auto prefetching :)

shiva-menta:
Yes that is a good idea, I'll prioritize that.

PR #614, opened 2024-04-04
Add PDP to GH workflow and k8s kittyhawk config

AaDalal:


AaDalal:
@esinx @joyliu-q Do you know how we actually deploy this to the domain we bought? (I think @Clue88 has details)

joyliu-q (commented):
Looks good to me! The domain should be auto configured because kittyhawk also manages Certificate orders. After CI/CD runs well can see if it is up and debug if not.

It looks like yarn check didn't run because it's not added to package.json though. Can y'all add the command?

AaDalal:
Passes all checks so per the slack convo good to merge.

PR #621, opened 2024-04-07
PDP Fixes

AaDalal:
Frontend
1. wrapping on search conditions
2. filter by startingYear when searching for degrees

Backend
1. update degree parser to handle course requirements where the department/disciplien is "@"
2. fixed a bug with the way the person field of docked course was being populated

Requires reloading degrees


Clue88 (approved):
LGTM

PR #624, opened 2024-04-07
Onboarding backend

ashleymyan:
• Added DegreeProfile and CourseTaken models to store onboarding info
• Basic CRUD routes, serializers, and unit tests
• Transcript model and transfer credit attribute are there but not yet implemented

### DIFF
--- a/backend/degree/admin.py
+++ b/backend/degree/admin.py
@@ -62,7 +62,6 @@ class DegreeAdmin(admin.ModelAdmin):
 
     def degree_editor(self, request):
         context = dict(self.admin_site.each_context(request))
-        return TemplateResponse(request, "degree-editor.html", context)
     
 
 class UserProfileAdmin(admin.ModelAdmin):
--- a/backend/degree/serializers.py
+++ b/backend/degree/serializers.py
@@ -30,6 +30,7 @@ from degree.models import (
 
 
 
+
 class DegreeListSerializer(serializers.ModelSerializer):
     class Meta:
         model = Degree
--- a/frontend/yarn.lock
+++ b/frontend/yarn.lock
@@ -4,7 +4,7 @@
 
 "@aashutoshrathi/word-wrap@^1.2.3":
   version "1.2.6"
-  resolved "https://registry.yarnpkg.com/@aashutoshrathi/word-wrap/-/word-wrap-1.2.6.tgz#bd9154aec9983f77b3a034ecaa015c2e4201f6cf"
+  resolved "https://registry.npmjs.org/@aashutoshrathi/word-wrap/-/word-wrap-1.2.6.tgz"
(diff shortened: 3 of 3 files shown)


backend/degree/models.py
```diff
@@ -549,6 +549,184 @@ class Meta:
...
+        max_length=5,
+        help_text=dedent(
+            """
+            The user's expected graduation date (of the form YYYYx where x is A [for spring],
+            B [summer], or C [fall]), e.g. `2019C` for fall 2019)
+            """
+        ),
+    )
+
+    degrees = models.ManyToManyField(
```
AaDalal, line 605: Why do we store degrees here? How would we display this on the onboarding panel.

backend/degree/models.py
```diff
@@ -549,6 +549,184 @@ class Meta:
...
+        through="CourseTaken",
+        related_name="degree_profile",
+        help_text=dedent(
+            """
+            A list of course codes that the user has already taken, matched with semester
+            """
+        ),
+    )
+
+    def calculate_total_credits(self):
```
AaDalal, line 635: Are we planning on using this anywhere?

backend/degree/views.py
```diff
@@ -189,3 +193,69 @@ def create(self, request, *args, **kwargs):
...
+    def get_serializer_class(self):
+        if self.request.method == "PATCH":
+            return DegreeProfilePatchSerializer
+        return super().get_serializer_class()
+
+    def perform_create(self, serializer):
+        user_profile, _ = UserProfile.objects.get_or_create(user=self.request.user)
+        return serializer.save(user_profile=user_profile.id)
+
+    def update(self, request, *args, **kwargs):
```
AaDalal, line 214: if we have a ModelViewset, the nice part is we don't have to define most of these things. See e.g., the SectionList view. This is because django rest framework ModelViewset already has defaults for these things.

backend/degree/views.py
```diff
@@ -189,3 +193,69 @@ def create(self, request, *args, **kwargs):
...
+
+        if instance.user_profile.user != request.user:
+            return Response(
+                {"detail": "Unable to delete profile."}, status=status.HTTP_403_FORBIDDEN
+            )
+
+        return super().destroy(request, *args, **kwargs)
+
+    @action(detail=True, methods=["post"], url_path="add-course", url_name="add_course")
+    def add_course(self, request, pk=None):
```
AaDalal, line 232: This is nice, I like this pattern for adding/removing on manytomany fields

AaDalal (changes requested):
Overall pretty great job using Django/Django rest framework concepts! I left some comments. I think the biggest thing to think about is what is the minimum information we need to store? And how can we use that on the frontend?

PR #640, opened 2024-04-09
add dnd support to search panel, highlight searched rule, UI adjustments

yuntongf:
- Can now drag a course from search result to plan panel, req panel, dock
- When user clicks search for a specific rule, the rule will be highlighted on the req panel as the search panel displays search results
- Reduce gap between req and search panels


frontend/degree-plan/components/FourYearPlanPage.tsx
```diff
@@ -104,6 +104,7 @@ const FourYearPlanPage = ({
   const ref = useRef(null);
 
   // search panel
+  const [searchedRuleId, setSearchedRuleId] = useState(-1);
```
AaDalal, line 107: search state is in a [context](https://github.com/pennlabs/penn-courses/blob/9d83d4dcab84f1d16924e2b7b669879fbf68219d/frontend/degree-plan/components/Search/SearchPanel.tsx#L20) since we set it from a bunch of different places in the UI... do we also need this here?

frontend/degree-plan/components/Requirements/QObject.tsx
```diff
@@ -72,14 +72,18 @@ const Attributes = ({ attributes }: { attributes: string[] }) => {
...
     flex-wrap: wrap;
     gap: .5rem;
     background-color: var(--primary-color-light);
     box-shadow: 0px 0px 14px 2px rgba(0, 0, 0, 0.05);
     cursor: pointer;
     padding: .5rem .75rem;
+    ${props => !!props.$isSearched && `
+        border-radius: 10px;
+        box-shadow: 8px 6px 10px 8px #00000026;
+    `}
```
AaDalal, line 86: I think this shadow looks a little weird bc it's so strong. Maybe we could change the color or apply some other effect to the surface of the search condition when clicked? let me know what you think

AaDalal (commented):
couple of questions

frontend/degree-plan/components/FourYearPlanPage.tsx
yuntongf: changed this to using searchRuleId which already exists instead of a new state variable

frontend/degree-plan/components/Requirements/QObject.tsx
yuntongf: I changed this to lighter and less aggressive shadow, similar to the one shown for droppable boxes

frontend/degree-plan/components/Search/SearchPanel.tsx
```diff
@@ -95,11 +100,17 @@ export const SearchPanel = ({ activeDegreeplanId }: { activeDegreeplanId: Degree
         setQueryString("");
     }, [ruleId])
 
+    const handleCloseSearch = () => {
+        setQueryString(""); 
+        setSearchPanelOpen(false);
+        setSearchedRuleId(-1);
```
AaDalal, line 106: I think the value should be null instead of -1

AaDalal (approved):
LGTM except 1 comment. Just gonna approve but if you could fix that and check ui after #650 is merged that would be amazing

frontend/degree-plan/components/Search/SearchPanel.tsx
yuntongf: Yea makes sense

PR #648, opened 2024-04-09
Add a dock clear button

JHawk0224:
Feel free to reject lol

PR #650, opened 2024-04-09
Fixes suggested by Jordan / Luke

AaDalal:
- [x] support for fulfilling with 3 digit codes
- [x] make the general course search more clear on the UI
- [x] make the root font-size smaller so the UI is less cramped
- [x] make year a free input field
- [x] make edit mode not turn off when you add or delete a semester
- [x] course credits are sometimes wrong. Two potential root causes:
  - [x] credits sql is wrong (e.g., MATH 1400 is not 2 credits)
  - [x] course with no credits defaults to 1 CU on the frontend. Need to make sure we get the course of that code with credits from the backend.

For a later pr:
- [ ] add clear course dock button -> need to make sure this doesn't lead to some weird state with fulfillments (jordan in #648)
- [ ] fix issues with radix ui select menus being partially hidden

### DIFF
--- a/backend/courses/management/commands/recompute_soft_state.py
+++ b/backend/courses/management/commands/recompute_soft_state.py
@@ -100,7 +100,7 @@ def recompute_enrollment():
 
 
 # course credits = sum(section credis for all activities for sections below 500)
-# the < 500 heuristic comes from here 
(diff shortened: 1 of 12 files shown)


JHawk0224:
❤️

PR #665, opened 2024-04-13
fix merge retry

yuntongf:


frontend/degree-plan/components/Search/CourseInSearch.tsx
```diff
@@ -171,10 +163,12 @@ export default function Course({
             <ReviewPanelTrigger full_code={course.id} triggerType="click">
                 <CourseContainer onMouseEnter={() => setIsMouseOver(true)} onMouseLeave={() => setIsMouseOver(false)}>
                     <CourseInfoContainer
+                        ref={drag} 
+                        className="draggable"
                         role="button"
                     >   
                         <CourseIdentityContainer >
-                            <CourseIDContainer ref={drag} className="draggable">
```
AaDalal, line 177: do we want to revert this? doesn't this allow the course to be draggable?

AaDalal, line 177: do we want to revert this? doesn't this allow the course to be draggable?

frontend/degree-plan/components/Search/SearchPanel.tsx
```diff
@@ -81,10 +81,10 @@ const SearchPanelHeader = styled(PanelHeader)`
 
 interface SearchPanelProp {
     activeDegreeplanId: DegreePlan["id"] | null;
-    setSearchedRuleId: (arg0: number) => void;
+    setSearchRuleId: (arg0: number | null) => void;
 }
 
-export const SearchPanel = ({ activeDegreeplanId, setSearchedRuleId }: SearchPanelProp) => {
+export const SearchPanel = ({ activeDegreeplanId, setSearchRuleId }: SearchPanelProp) => {
```
AaDalal, line 87: why do we need to explicitly pass this down as a prop? could we just get it from the context?

frontend/degree-plan/components/Search/SearchPanel.tsx
```diff
@@ -127,6 +127,7 @@ export const SearchPanel = ({ activeDegreeplanId, setSearchedRuleId }: SearchPan
                         value={queryString}
                         onChange={(e) => {setQueryString(e.target.value)}}
                         autoComplete="off"
+                        // placeholder={!ruleId ? "Search for a course!" : `Filtering for ${ruleQuery ? ruleQuery : 'a requirement'}`}
```
AaDalal, line 130: nit: maybe we can get rid of this?

AaDalal (changes requested):
mostly looks good except a few comments that would be great to fix before we merge :^). LMK when you've fixed and I can stamp

frontend/degree-plan/components/Search/SearchPanel.tsx
AaDalal: context is used on the next few lines

frontend/degree-plan/components/Search/CourseInSearch.tsx
yuntongf: Oh this is because I put the ref in the parent CourseInfoContainer component, but I think putting it in the CourseIDContainer component makes more sense in terms of UI. I'll change it.

frontend/degree-plan/components/Search/SearchPanel.tsx
yuntongf: Yea legacy code I will delete it

frontend/degree-plan/components/Search/CourseInSearch.tsx
yuntongf: Actually I'll put it in CourseIdentityContainer

AaDalal: Oh oops missed that it was on the parent. Do whatever makes sense to you!

yuntongf:
Should I also merge the changes from aug/add-semester-popup? In that branch I changed the add semester thing to the top of the plan panel and a modal will pop up on click

AaDalal:
Up to you -- im good either way. Let's just do that as a merge rather than a rebase so we can always go back to this commit if needed.

PR #667, opened 2024-04-24
Add Status Audit Management Command

shiva-menta:
Adding a management command that generates a text file, displaying the differences in section statuses between our database and the OpenData API endpoint `/v1/course_section_status/{term}/all`.

### DIFF
--- a/backend/alert/management/commands/statusaudit.py
+++ b/backend/alert/management/commands/statusaudit.py
@@ -1,0 +1,98 @@
+import json
+import logging
+
+from django.core.management.base import BaseCommand
+from tqdm import tqdm
+
+from alert.models import Course, Section
+from courses import registrar
+from courses.util import get_course_and_section, get_current_semester, translate_semester_inv
+
+
+class Command(BaseCommand):
+    help = """
+        Generate an audit report that demonstrates the differences between our
+        database and the course statuses received from using the OpenData
+        endpoint directly.
+
+        Note that this script DOES NOT make any changes to the database, just
+        generates a textfile report
+    """
+
+    def handle(self, *args, **options):
+        root_logger = logging.getLogger("")
+        root_logger.setLevel(logging.DEBUG)
+
+        semester = get_current_semester()
+        statuses = registrar.get_all_course_status(semester)
+        stats = {
+            "missing_data": 0,
+            "section_not_found": 0,
+            "duplicate_updates": 0,
+            "unsynced_updates": 0,
+        }
+        unsynced_courses = []
+        for status in tqdm(statuses):
+            data = status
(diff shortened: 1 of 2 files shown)


backend/alert/management/commands/statusaudit.py
```diff
@@ -0,0 +1,98 @@
...
+                stats["duplicate_updates"] += 1
+                continue
+
+            stats["unsynced_updates"] += 1
+            unsynced_courses.append(
+                (section_code, last_status_update.new_status, current_status, course_status)
+            )
+
+        # Write out statistics and missing courses to an output file.
+        with open("./status_audit.txt", "w") as f:
```
AaDalal, line 74: if we want to run this on a cron, we may want to allow the file name to be passed in/to include the UNIX timestamp.

backend/alert/management/commands/statusaudit.py
```diff
@@ -0,0 +1,98 @@
...
+            f.write("Summary Statistics\n")
+            f.write(json.dumps(stats) + "\n\n")
+
+            f.write(
+                """Courses Out of Sync\nCourse Code / Last Update Status /
+                 Our Stored Status / Actual Status\n"""
```
AaDalal, line 80: What does actual status mean? Might be good to elaborate bc I'm getting it mixed up in my head

backend/alert/management/commands/statusaudit.py
```diff
@@ -0,0 +1,98 @@
...
+        }
+        unsynced_courses = []
+        for status in tqdm(statuses):
+            data = status
+            section_code = data.get("section_id_normalized")
+            if section_code is None:
+                stats["missing_data"] += 1
+                continue
+
+            course_status = data.get("status")
```
AaDalal, line 42: is this status per section or per course?

AaDalal (commented):
lgtm, mostly just nits

backend/alert/management/commands/statusaudit.py
shiva-menta: Yup sounds good. Do you think we should write it to S3 or something like that? Maybe it'd be cool if we could setup a Slack integration and get notifs that way, but might be a lot of extra work for little reward.

backend/alert/management/commands/statusaudit.py
shiva-menta: Actual status is the Path@Penn (or the accurate), status of a course. I'll rename it a bit and also add a comment explaining in-line.

backend/alert/management/commands/statusaudit.py
shiva-menta: It should be `section_status`, will change.

PR #668, opened 2024-04-24
PCA Resyncing Logic

shiva-menta:
- Adding `transaction.atomic()` wrapper to webhook database calls, to reduce potential consistency errors between our StatusUpdate models and our Section model. This required modifying one of our existing test cases.
- Modifying logic for `loadstatus` management command to conditionally add StatusUpdate objects to ensure consistency between the last StatusUpdate and the status attribute of the Section model. `registrarimport` also uses `loadstatus`'s logic, but it currently does not add StatusUpdate objects – this can be changed easily by setting `add_status_update=True`.
- Add `sync_path_status` management command to sync our database with path statuses in the case that OpenData's status API is inaccurate (we've historically seen this for around 4% of courses).


backend/tests/alert/test_alert.py
```diff
@@ -806,9 +806,7 @@ def test_alert_called_wrong_sem(self, mock_alert):
             "sent" in json.loads(res.content)["message"],
         )
         self.assertFalse(mock_alert.called)
-        self.assertEqual(1, StatusUpdate.objects.count())
-        u = StatusUpdate.objects.get()
-        self.assertFalse(u.alert_sent)
```
AaDalal, line 811: why does this test change?

backend/courses/management/commands/loadstatus.py
```diff
@@ -19,22 +25,56 @@ def set_all_status(semester=None):
...
+        last_status_update = section.last_status_update
+        current_status = section.status
+
+        # Change status attribute of section model (might want to use bulk update)
+        if current_status != course_status:
+            section.status = course_status
+            section.save()
+
+        # Add corresponding status update object
+        if add_status_update and last_status_update.new_status != course_status:
```
AaDalal, line 54: This does resyncing manually, but this means we would have to run this on a corn all the time right? This definitely solves one problem, but we still probably need to do something to fix the underlying issue (why do they mismatch ever?)

AaDalal (approved):
mostly LGTM, but left some comments I would appreciate if you could look over before you merge

backend/tests/alert/test_alert.py
shiva-menta: My understanding is this test was kind of flawed from the beginning – it attempts to create a StatusUpdate object with a poorly formatted semester. In the `record_update` function there is a check after the creation of the object that raises an error if the semester is poorly formatted. When changing it to `transaction.atomic()`, throwing this error leads to reverting the creation of StatusUpdate, which makes more sense logically anyways, as a StatusUpdate object with a poorly formatted semester doesn't make sense.

PR #674, opened 2024-09-13
Fix docs override AutoSchema

el-agua:
- creates new model_field fn that removes recursive calling to model_serializer fn

### DIFF
--- a/backend/PennCourses/docs_settings.py
+++ b/backend/PennCourses/docs_settings.py
@@ -6,12 +6,15 @@ from inspect import getdoc
 from textwrap import dedent
 
 import jsonref
+from django.db import models
 from django.urls import get_resolver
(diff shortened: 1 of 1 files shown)


backend/PennCourses/docs_settings.py
```diff
@@ -1012,6 +1016,130 @@ def map_serializer(self, serializer):
 
         return result
 
+    # Overrides, uses overridden method
+    # (https://www.django-rest-framework.org/api-guide/schemas/#map_serializer)
```
shiva-menta, line 1020: nit: change this to https://www.django-rest-framework.org/api-guide/schemas/#map_field

shiva-menta (approved):
lgtm - thanks for the work!

PR #679, opened 2024-10-22
Deprecate Dev Dockerfile, Update README, Relock Pipfile, Add psql to Dev Container

shiva-menta:
Various environment changes to reflect the current state of PCX backend:
- Deprecating Dev Dockerfile – Abandoning this in favor of developing in Dev Containers. Old `Dockerfile.dev` also was pinned to an old version of Python.
- Updating README – Removing instructions related to `Dockerfile.dev`, adding Dev Container-specific instructions, and cleaning up the content a bit.
- Relocking Pipfile – Resolving lingering versioning issues when running the backend for the first time.
- Add psql to Dev Container – We previously didn't install `postgresql-client` within the Dev Container, but this is needed for loading PCX data into database. Based on Dev Container `ubuntu` version, `psql` is outdated at version `14` compared to new pinned version of postgres at version `15.8` (closest version to our current prod database), but this should be okay.


el-agua (approved):
lgtm

PR #680, opened 2024-10-24
made pennkeys lowercase and changed test case

nruia-penn:
Made it so when people send friend requests on penn courses, it changes the username to all lowercase so that when people add friends and accidentally capitalize the first letter it still is able to find the friend they are looking for.

### DIFF
--- a/backend/courses/views.py
+++ b/backend/courses/views.py
@@ -521,7 +521,6 @@ class FriendshipView(generics.ListAPIView):
 
     def post(self, request):
         sender = request.user
-        res = {}
 
         username = request.data.get("pennkey")
         if not username:
@@ -563,7 +562,6 @@ class FriendshipView(generics.ListAPIView):
         # (depends on who sends the request)
 
         sender = request.user
-
         res = {}
 
         username = request.data.get("pennkey")
--- a/backend/courses/views.py
+++ b/backend/courses/views.py
@@ -520,8 +520,14 @@ class FriendshipView(generics.ListAPIView):
 
     def post(self, request):
         sender = request.user
+        res = {}
 
-        recipient = get_object_or_404(User, username=request.data.get("pennkey").lower())
+        username = request.data.get("pennkey")
+        if not username:
+            res["message"] = "User not found"
+            return Response(res, status=status.HTTP_404_NOT_FOUND)
+
+        recipient = get_object_or_404(User, username = username.lower())
 
         existing_friendship = (
             self.get_all_friendships().filter(Q(recipient=recipient) | Q(sender=recipient)).first()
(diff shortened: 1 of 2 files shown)


backend/courses/views.py
```diff
@@ -520,7 +520,8 @@ def get_all_friendships(self):
 
     def post(self, request):
         sender = request.user
-        recipient = get_object_or_404(User, username=request.data.get("pennkey"))
+
+        recipient = get_object_or_404(User, username=request.data.get("pennkey").lower())
```
Clue88, line 524: You should probably build in some logic here to handle the case where `request.data.get("pennkey")` returns `None`; this probably shouldn't happen based on input validation but we wouldn't want this to crash because of NoneType exception. (same with the other case that you modified)

You might also want to add some tests for unexpected inputs if they're not already there

Clue88 (changes requested):
Looks good! Left a comment for some edge cases

shiva-menta:
> Just made all pennkeys lowercase. Made a test for it, seems to work.

Would also be great to add 1-2 sentences on why this is needed! It's helpful to have more documentation on why we're making changes.

nruia-penn:
Do you want me to add comments in the code or in some sort of documentation?

shiva-menta:
> Do you want me to add comments in the code or in some sort of documentation?

Just within the PR itself would be sufficient. Something like [this](https://github.com/pennlabs/penn-courses/pull/667).

backend/courses/views.py
```diff
@@ -520,7 +521,14 @@ def get_all_friendships(self):
 
     def post(self, request):
         sender = request.user
-        recipient = get_object_or_404(User, username=request.data.get("pennkey"))
+        res = {}
+
+        username = request.data.get("pennkey")
+        if not username:
+            res["message"] = "User not found"
+            return Response(res, status=status.HTTP_404_NOT_FOUND)
```
shiva-menta, line 529: Can make this more succinct and use `raise Http404("User not found.")`.

backend/courses/views.py
```diff
@@ -554,9 +562,17 @@ def post(self, request):
...
+
         sender = request.user
-        recipient = get_object_or_404(User, username=request.data.get("pennkey"))
+
+        res = {}
+
+        username = request.data.get("pennkey")
+        if not username:
+            res["message"] = "User not found."
+            return Response(res, status=status.HTTP_404_NOT_FOUND)
```
shiva-menta, line 573: Same thing here.

shiva-menta (approved):
lgtm – thanks Nikhil!

Clue88 (approved):
looks great!

PR #682, opened 2024-11-05
feat(webhook): PCA 2.0 webhook

esinx:


esinx:
goodbye bun you don't work really well with lambda and api gateway, I'll come back to you once you're actually compatible with API Gateway events

PR #685, opened 2024-11-13
Fix Precompute Topics Script for Manually Set Parent Courses

shiva-menta:
This semester, we saw two major course renumbers in the CIS department – 3800 to 4480 and 3410 to 4521. To quickly modify our database so searches to the new course numbers show reviews from previous numbers, I added the `2024A` offerings of these courses as the parent course for 4480 and 4521 in `2025A`, as well as setting `manually_set_parent_course` to `True`.

While doing this, I noticed some issues with our `recompute_topics` script, which needed to be resolved in order for these course links to show up properly:
- Added a `visited_courses` set to make sure each course is assigned only one topic.
- Added the following role to our topic computation logic: Any child with siblings that is the only child with a manually set `parent_course_id` inherits the parent topic.

After updating the DB and rerunning topic computation, we get the following:
Screenshot 2024-11-12 at 11 29 30 PM
Screenshot 2024-11-12 at 11 29 45 PM

Also adding `batch_size` for review precomputation.

### DIFF
--- a/backend/courses/management/commands/recompute_topics.py
+++ b/backend/courses/management/commands/recompute_topics.py
@@ -53,11 +53,16 @@ def recompute_topics(min_semester: str = None, verbose=False, allow_null_parent_
                 Course.objects.filter(semester=semester)
                 .select_related("parent_course", "parent_course__topic")
                 .prefetch_related("parent_course__children")
+                .order_by("-manually_set_parent_course")
             )
             topics_to_create = []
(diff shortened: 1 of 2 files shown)


Clue88 (approved):
thanks shiva! lgtm👍

el-agua (approved):

nruia-penn (approved):
seems to make sense

PR #698, opened 2025-01-17
Reduce sync interval

anli5005:
As part of our PCA spam mitigation plan, this reduces the sync interval.


anli5005:
Closing since this dupes #697

PR #703, opened 2025-02-16
Break CRUD routes

nruia-penn:
This pr adds a break view set so that someone can create and update breaks on PCP in the backend.

This is a simple view set with CRUD routes that should let a user add breaks to their schedule in penn course plan.

I wrote test cases in test_schedule.py.

Some design decisions: 
- I changed the meeting model to have either an associated Section or associated break but never both, this is so breaks can also have a location. If we choose not to use location (this could be hard to implement on front end with how many rooms there are in this school), there is also an option for a simple location_string, which can be set by the user.
- The break model was put in plan/models.py not plan/courses.py, this makes more sense to me as it is a model only used by PCP.


el-agua (changes requested):
Good work! Just left a few comments to look at.

el-agua (changes requested):
Small thing, but looking good!

nruia-penn:
I think work now?

luke-rt:
Does not look good to me

PR #716, opened 2025-04-23
Fixes server error for OpenAPI

el-agua:


el-agua:
> how does this fix server error for OpenAPI?
> 
> but looks good

Not sure entirely, but docs have always been fine on local. That particular line throws a sentry 500 error on prod so OpenAPI won’t render. And in general, eval bad lol

PR #718, opened 2025-05-14
PCP Breaks

el-agua:
PCP Breaks

Frontend Changes (@olivianhu):
- Fixed mobile tab view to include previous features.
- Added breaks tab.
- Created new reducers to facilitate breaks CRUD
- Update backend sync to account for deleting/creating/toggling breaks

Backend Changes (@nruia-penn @el-agua):
- Added new Breaks model and respective views
- Meeting objects can now belong to breaks and sections


luke-rt:
Frontend
Important stuff
- Maybe make the slider increment by 5 minutes? or some other solution where you can type in the time of the break so setting break times is easier to use.

More nitpicky design stuff
- The input for break name I think doesn't really match the design system, maybe make it look like the course search or the input for schedule name.
- Checkboxes for what day the break is on should maybe be PCP purple

luke-rt:
Looks good, just one small thing maybe remove the blue outline when typing in the Break name input box since the schedule input doesn't have it. Other than that im down to merge

luke-rt (approved):
Nice 👍

PR #694, opened 2024-11-22
Add Status Updates and Cancelled Courses

shiva-menta:
- Making sure to add StatusUpdate object for daily `loadstatus` command.
- Add cancelled course support for Path@Penn API.
- Change Path sync frequency to 30 minutes.

PR #720, opened 2025-07-20
Removed null pointer on status change

el-agua:
- On script that changes statuses, considers the case where there does not exist a "last status"
- Moves semaphores to non-global (locking issue)

### DIFF
--- a/backend/courses/management/commands/loadstatus.py
+++ b/backend/courses/management/commands/loadstatus.py
@@ -65,16 +65,35 @@ def set_all_status(semester=None, add_status_update=False, verbose=False):
             sections_to_update.append(section)
             statuses_out_of_sync.append(section_code)
 
-        if add_status_update and last_status_update.new_status != course_status:
-            record_update(
-                section,
-                course_term,
-                last_status_update.new_status,
-                course_status,
-                False,
-                json.dumps(status),
-            )
-            status_updates_out_of_sync.append(section_code)
+        if add_status_update:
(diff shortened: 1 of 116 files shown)


backend/courses/management/commands/loadstatus.py
```diff
@@ -65,16 +65,35 @@ def set_all_status(semester=None, add_status_update=False, verbose=False):
...
-                json.dumps(status),
-            )
-            status_updates_out_of_sync.append(section_code)
+        if add_status_update:
+            if not last_status_update and course_status:
+                try:
+                    record_update(
+                        section,
+                        course_term,
+                        "",
```
jonathanmelitski, line 74: Is there a default status that should exist instead of empty string? Not exactly sure what the existing workflow here but it seems like we should be careful with the data that we write?

el-agua: "" is the key for unlisted, see courses/models.py (Section.STATUS_CHOICES), added a comment to make this more explicit

el-agua:
After it is deployed, should just be run by the cron job. There were some changes last fall (https://github.com/pennlabs/penn-courses/pull/694) to change how alerts were sent out, and I think it was just missing a null check when a new semester was loaded in.

jonathanmelitski (approved):
Giving it the stamp, but you should probably run black locally so that it passes CI.

Thanks for doing this. Also yeah completely my bad for not understanding what that script was doing. Thanks for clarifying.

PR #733, opened 2025-09-28
Fix test cases

luke-rt:
Fix test cases
- Made JwtAuthentication return 403 when it fails
- Fixed test cases to match the updated autocomplete route


olivianhu (approved):
LGTM!

el-agua (approved):

PR #737, opened 2025-10-05
Andrew k newbie project

i30101:
Timebar to show current time in the schedule grid, doesn't show when day is in the weekend and the weekend isn't visible in the schedule grid


i30101:
Screenshot 2025-10-05 at 2 15 46 PM

PR #738, opened 2025-10-05
Fix remove break error

i30101:
Removing a break from the schedule grid gave an error, fixed now


frontend/plan/components/schedule/ScheduleDisplay.tsx
jamesdoh0109: Nit: can we also use enum here instead of raw string so we are consistent

jamesdoh0109: Why cast to Weekdays?

frontend/degree-plan/tsconfig.tsbuildinfo
jamesdoh0109: I would not commit this

PR #770, opened 2026-02-15
Adds caching for review data for all PCx products

el-agua:
Instead of calculating review averages, precomputed with the cache_course_reviews django command. Will be calculated manually. 

1. Test fallback by checking if reviews populate WITHOUT running the command
2. Run cache_course_reviews management command
3. Check review scores populate and latency is low


samuel-lao:
gg so fast

samuel-lao (approved):

PR #773, opened 2026-02-21
Create 0069_userprofile_has_been_onboarded.py

el-agua:


ryantanen (changes requested):
I wonder if this is even a smart column to add to user profile. What if we eventually have more than one onboarding flow (i.e penn course plan vs penn degree plan onboarding?). Would better design be to have a separate UserOnboarding Table?

If we do this I'll be monitoring for any long running queries n user profile, as I believe this takes an exclusive lock

PR #792, opened 2026-09-11
Add submatriculation support

ronitkapoor0106:
Backend — Path lists masters programs in the same catalog under degree code MSE, all reporting program code EM_MSE, and their audits parse exactly like a bachelors (DEGREE + MAJOR block, plus CONC when concentrated). fetch_degrees_from_path now fetches them alongside the undergraduate programs, with --skip-masters to opt out. A full scrape of --srcdb 202630 saves 44 MSE degrees across 12 SEAS majors.

Frontend — A transcript holds one academic record per Program: line. The parser now splits on those and scans each separately, so a record's majors stay paired with its own concentrations and match against its own school's degrees. A masters record fills a new Submatriculation field, which posts alongside the bachelors so a course on both records counts toward rules in both degrees.

Testing — 5 new unit tests on the program filtering (mutation-tested); 147 backend tests pass. Verified end to end against a real submat transcript: both degrees attach to one plan


backend/degree/utils/path_client.py
```diff
@@ -39,6 +39,11 @@
 # minors and certificates from the same catalog, which we skip.
 UNDERGRADUATE_DEGREE_CODES = {"BA", "BAS", "BS", "BSE", "BSN"}
 
+# Graduate degree codes we model, for submatriculants pursuing a masters alongside their
+# bachelors. Path lists many more graduate degrees (MBA, PHD, MD, ...) that we do not model;
+# only SEAS's MSE is submat-eligible in a way PDP can currently represent.
+MASTERS_DEGREE_CODES = {"MSE"}
```
el-agua, line 45: If we have the information, I would like a masters degree to be first-class supported like undergraduate degrees. (e.g. we should support people who are ONLY masters students)

frontend/degree-plan/components/OnboardingPanels/CreateWithTranscriptPanel.tsx
```diff
@@ -45,6 +45,7 @@ import { postFetcher, getCsrf } from "@/hooks/swrcrud";
 import {
   getMajorOptions,
   getSecondMajorOptions,
+  getSubmatOptions,
```
el-agua: Same point as above. Undergrad and graduate degrees should both live together.

This will probably reduce some duplication on the UI code

frontend/degree-plan/components/OnboardingPanels/CreateWithTranscriptPanel.tsx
```diff
@@ -163,8 +167,10 @@ export default function CreateWithTranscriptPanel({
             JSON.stringify(semesters)
           );
         }
+        // A submatriculant's masters is a degree on the same plan as their bachelors, which is
+        // what lets a shared course count toward rules in both.
```
el-agua: I would be careful for this. From my knowledge, path at penn does not AUTOMATICALLY double count things between undergrad and masters.

el-agua (changes requested):

backend/degree/utils/path_client.py
ronitkapoor0106: masters go through the same Majors field as any other degree now

frontend/degree-plan/components/OnboardingPanels/CreateWithTranscriptPanel.tsx
ronitkapoor0106: yea its one drop down now

frontend/degree-plan/components/OnboardingPanels/CreateWithTranscriptPanel.tsx
ronitkapoor0106: yeah i fixed so only the first 3 courses that qualify for the masters are what's double counted

PR #794, opened 2026-09-11
Backend pcr-redesign

el-agua:
- Add a `departments` query param to CourseSearchFilterBackend that filters by pipe-separated department codes.
- Enable OptionalPageNumberPagination (capped at 500 per page) and CourseSearchFilterBackend on the base /courses/ list endpoint.
- When the "all" semester is requested with section-specific filters (days, time, instructor_quality), snap to the current semester.

### DIFF
--- a/backend/courses/views.py
+++ b/backend/courses/views.py
@@ -76,7 +76,8 @@ class BaseCourseMixin(AutoPrefetchViewSetMixin, generics.GenericAPIView):
             queryset = queryset.filter(**{self.get_semester_field(): semester})
         else:  # Only used for Penn Degree Plan (as of 4/10/2024)
             queryset = (
-                queryset.exclude(credits=None)  # heuristic: if the credits are empty, then ignore
+                # heuristic: if the credits are empty, then ignore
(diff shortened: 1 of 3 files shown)


backend/courses/views.py
```diff
@@ -178,8 +179,24 @@ class CourseList(generics.ListAPIView, BaseCourseMixin):
...
+    filter_backends = [CourseSearchFilterBackend]
     queryset = Course.objects.none()  # included redundantly for docs
 
+    # These params filter on schedule/section data that only exists in a specific semester,
+    # so requesting them on the "all" semester path automatically snaps to current semester.
+    _SEMESTER_SPECIFIC_PARAMS = frozenset({"days", "time", "instructor_quality"})
+
+    def get_semester(self):
+        semester = super().get_semester()
+        if semester == "all" and self._SEMESTER_SPECIFIC_PARAMS.intersection(
```
el-agua: In general, we should probably make this representation illegal (callers will get confused). All + semester_specific_params doesn't really make sense...., maybe we should add a `latest` slug

